PR #2, opened 2022-12-01
updated messaging.wit.md, and README.md

danbugs:
Signed-off-by: danbugs [danilochiarlone@gmail.com](mailto:danilochiarlone@gmail.com)


messaging.wit.md
```diff
@@ -0,0 +1,105 @@
...
+    }
+    
+    /// An error type
+    type error = string
+}
+
+/// An interface for a producer.
+interface "wasi:messaging/pub" { 
+	/// Publishes an event to a channel.
+	publish: func(c: channel, e: event) -> result<_, error>
```
stevelr, line 51: should this be a topic string, rather than a channel? Isn't the choice of queue vs topic a consumer issue?

messaging.wit.md
```diff
@@ -0,0 +1,105 @@
...
+    	specversion: string,
+    	ty: string,
+    	source: string,
+    	id: string,
+    	data: option<list<u8>>,
+    	datacontenttype: option<string>,
+    	dataschema: option<string>,
+    	subject: option<string>,
+    	time: option<string>,
+    }
```
stevelr, line 22: I really like that this uses the CloudEvents standard!
How could we add support for [extension attributes](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#extension-context-attributes) ?
Should there be another field `extensions: option<map<string,string>>` ?

messaging.wit.md
```diff
@@ -0,0 +1,105 @@
...
+	/// Creates an on-receive handler push-based message delivery.
+  	on-receive: func(e: event) -> result<_, error>
+}
+
+/// An interface for a consumer relying on basic pull-based message delivery.
+interface "wasi:messaging/basic/pull" {
+	/// Pulls a message.
+  	receive: func() -> result<event, error>
+}
+
```
stevelr: does the result generate an ack, or should there be a separate method for that? 

For cases where a receiver would ack and continue processing, the ack might need to be a separate method. (since wasm is single threaded, it can't spawn an ack) _or_ we need some kind of deferred callback, which is more complicated.

messaging.wit.md
```diff
@@ -0,0 +1,105 @@
...
+/// An interface for a consumer relying on push-based message delivery.
+interface "wasi:messaging/push" {
+	/// Creates an on-receive handler push-based message delivery.
+  	on-receive: func(e: event) -> result<_, error>
+}
+
+/// An interface for a consumer relying on basic pull-based message delivery.
+interface "wasi:messaging/basic/pull" {
+	/// Pulls a message.
```
stevelr: in wasmCloud, a majority use case is `receive_timeout()`, which returns `result<event,Error>` where Error can be a timeout error or some other error. In this interface, the timeout option is probably applicable to both pull and push consumers.

stevelr (changes requested):

messaging.wit.md
danbugs: Thank you for bringing this up~ I'll make a commit including that!

messaging.wit.md
danbugs: `receive` would generate an ack — If the user wants to continue processing, I'd assume they'd want `stream-receive`. Although, that, in itself, might need a `commit` function or something of the sort to `ack`...

messaging.wit.md
danbugs: That's right~ I actually meant to include a `time-to-live` parameter for `receive` but it ended up escaping me. That said, I'm not sure that is applicable for push-based message delivery as there's no waiting from the user side.

messaging.wit.md
danbugs: That's a good point~ My initial thought was that, having, on the host side, the type of messaging infrastructure the consumer is using, would be helpful for the implementors of the interface to discern different ways of handling data.

For example, say we want to create an implementation of this interface for AWS' services: SQS, and SNS — on one side, SQS offers queues, and, on the other, SNS offers topics. With the distinction between topics and queues at the interface level, we could create a single implementation for AWS covering both competing and non-competing consumer patterns. If not, the implementations would have to be separate, and, consequently, not every implementor of this interface would provide the same behaviour. 

That said, I think it's not all that unreasonable that, if a user of this interface chooses an implementer of SQS, they would expect different behaviour...

stevelr: I haven't used SQS in a long time. In Nats, you publish to a topic, and the client/consumer chooses whether they want a queue (pool) subscription or a regular subscription (all messages go to all consumers). In NATS, the publisher doesn't get a choice. I think of it as a consumer implementation choice

stevelr: but if SQS doesn't give that choice, then I see how the channel parameter is needed

danbugs:
From the discussions in the [WASI Meeting from Dec. 1st 2022](https://github.com/WebAssembly/meetings/blob/main/wasi/2022/WASI-12-15.md), a couple of questions came up:
- Should we surface quality-of-service? With that, should we have an offset interface to allow exactly once delivery?
- Should we be called `wasi-pubsub` or `wasi-messaging`?
- Should we get rid of the `wasi-message-queue` proposal and focus and unify the concepts through this one?
- Should `on-receive`, and `stream-receive` also have a `time-to-live-in-milliseconds`?
- Should we have `ack()` and `on-ack()` methods?

messaging.wit.md
danbugs: Yeah~ It doesn't give you that choice. You go for SQS if you want queues, and you go for SNS if you want topics. Those two could be different implementors of the interfaces, but then they wouldn't provide the same functionality...

messaging.wit.md
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/// Two typical usage examples.
+world "wasi:messaging/pull-pubsub" {
+    import pub: "wasi:messaging/pub"
+    import sub-pull: "wasi:messaging/basic/sub-pull" // or "wasi:messaging/stream/sub-pull"
+  
+    export handle-receive: "wasi:http/handler"
+    export handle-subscribe: "wasi:http/handler"
+    export handle-publish: "wasi:http/handler"
```
Mossaka: I don't think "wasi:http/handler" exposes these functions. Do you mean "export handle-http: 'wasi:http/handler'"?

danbugs: We don't really have a "wasi:http/handler" defined, but — yes, you're right. My idea here is to convery exporting "named" handlers.

lukewagner:
Hi, great work on this!

High-level question: are the `world`s in the "World Examples" section of `messaging.wit.md` just example worlds that you *could* write, but aren't being standardized, or are these worlds being standardized alongside the interfaces in the "Interfaces" section?  If the former, maybe replace the `"wasi:messaging/*"` in their names with an unofficial-sounding kebab-name like `my-world` so it's a bit more clear.  If the latter, maybe remove "Examples" from the section title.

Another question, which may just be my lack of understanding of how these messaging systems work, so sorry in advance for that, but I'm having a hard time wrapping my head around the exported `pull` interface of the `basic/sub-pull` and `stream/sub-pull` worlds.  Is the idea that the host calls `receive` and then kindof just waits for an event, and this gives the component the chance to run to produce events?  If the timeout expires and `receive` returns, what causes the host to call `receive` again in the future (to let the component run again)?  Does it happen continuously or periodically or does the call to `receive` reflect some external event?  Overall, this feels sortof incompatible with the event-based/"serverless" model where the way this would work is that some *independent* event fires (HTTP request arrives, timer fires, message arrives via `push`, ...), and then *in response to that*, the component calls an imported `push`, `pub` or `sub` function.  Thus, I wouldn't imagine there even *being* a `pull` interface.  But maybe `receive` represents some external event, analogous to these other ones I listed?

Lastly, a few small ideas/nits:
* To allow efficient batch-processing of events via the `push` interface, instead of having a separate interface with a `stream` function, I'd recommend instead adding a second streaming function to `wasi:messaging/push` that has a `stream` param type.
* It might be nice to rename `wasi:messaging/push`/`on-receive` to `wasi:messaging/handler`/`handle` for symmetry with `wasi:http/handler` (they're both serving the same purpose, just different message types)?
*  One world I'd expect to see standardized but isn't quite is:
```wit
world "wasi:messaging/service" {
  import pub: "wasi:messaging/pub"
  import sub: "wasi:messaging/sub"
  export message-handler: "wasi:messaging/handler"
}
```

Mossaka:
Hey @lukewagner , good question! I will leave the first question and the last section of ideas/nits to @danbugs . I will try my bset to answer the second question regarding the `pull` interface. 

Before answer the quesitons, I want to give some background on how message delivery in a messaging system works. From a high level point of view, there are two major message delivery mechanisms between a message broker and a subscriber:`push` and `pull`. 

The `push` mechanism allows the host to continuously push messages to the subscriber component to handle. This will mostly match the serverless / function-as-a-service programming model, where the component only need to export a message handler function for the host to invoke. 

```rust
on_message( event: Event ) -> Result<...>
```

The `on_message` will be implemented by the component, exported, and invoked by the host.

The second message delivery mechanism is the `pull` mechanism. The idea is that the subscriber service is a **long running service** that periodically pulls / polls the broker to get messages. This actually gives the subscriber service more flexibility and useful if the subscriber has limited resources to process messages. 

```rust
receive(...) -> Result<Event>
```

The `receive()` function will be implemented by the host, imported and invoked by the guest. 

You are right that the `pull` mechanism does not feel right to a serverless model, but it is a wildly used mechanism for subscribers to receive messages from the message broker. It is working much more like an outbound HTTP request, though the request is not necessarily a http reqeust.

lukewagner:
Thanks a lot for the explanation @Mossaka, that makes sense.  The remaining detail I don't understand is why the `pull` interface is *exported* by the `sub-pull` world instead of being *imported*: for long-running instances, I would expect the export to be something like a `main` function that didn't take any event-specific parameters.

Next, an observation and a question: I definitely agree that it's important to be able to run code that assumes a traditional long-running-instance model.  However, it does seem like there is a way to support this code in the following manner that is also rather serverless- and autoscaling-friendly:
* The component's `start`/[`start2`](https://github.com/WebAssembly/component-model/issues/146) function executes the code's `main` entry up to the point where the code enters the (often framework-defined) main event loop, and returns.
  * If hosts do the snapshotting optimization discussed in [component-model/#146](https://github.com/WebAssembly/component-model/issues/146), then this execution gets factored out of runtime.
* When an external event arrives, the host instantiates and calls an event-specific export of the component which resumes the event loop and the component continues to execute until that event (and any I/O spawned by that event) completes and then, instead of blocking in the event loop, returns from the export.
  * If the underlying host is implemented in terms of a pull model, then the real blocking "pull" would happen in the host (allowing efficient multiplexing across many guest instances), passing the pulled messages as parameters to export calls.
* The host can continue to call exports on the same instance for as many subsequent events as it wants. When there are no pending events, the host can tear down the instance, keep the instance alive waiting for a future event, or some hybrid strategy using timeouts.

What's nice about this approach is that it avoids having two differently-shaped worlds (for serverless and long-running instances) and a corresponding bifurcation between hosts and components: there's just one world (for a particular set of functionality) and multiple ways to target and support it.

An obvious challenge with this approach is if the long-running guest code doesn't have an event loop but, rather, has non-concurrent, synchronous blocking `pull` calls.  This code can be supported using either core wasm [stack-switching](https://github.com/WebAssembly/stack-switching) or the component model's proposed [blocking `canon wait`](https://docs.google.com/presentation/d/1MNVOZ8hdofO3tI0szg_i-Yoy0N2QPU2C--LzVuoGSlE/edit#slide=id.g11b88b15c38_0_109).  Neither are available today or very soon, though, so that is a downside.  But with either of these, such synchronous code can be suspended and resumed mid-stack.

Given all that, do you still see a motivation for defining separate interfaces for long-running instances?  I can imagine hypothetical use cases where code might critically depend on having long-running instances (no autoscaling allowed) where a distinct blocking export interface would be necessary to reflect this requirement and, in this case, the component would want, e.g., to import a blocking `pull` (and probably other interfaces too).  But if we don't have those use cases in the short-term, maybe we can punt on them for now and benefit from the increased portability resulting from fewer degrees of freedom?

devigned:
I agree that the push model is the most natural way to go here, but in the same frame, I can see the desire for non-function kind of folks to depend on it. 

I believe the idea of a host holding a queue of messages for the application is a solid one, but would require some elegant work to deal with back pressure and configuration of the message broker, which would ordinarily fall on the consuming application.

In the first edition, rather than making on omnipresent app host, I'd rather suggest cutting pull. I think this would satisfy much of the community and cause the pull folks to speak up, which would drive a better use case from folks who have stake in the outcome.

wdyt?

danbugs:
Thanks for the feedback, @lukewagner, and @devigned!~ and thanks @Mossaka for addressing Luke's questions, your answers were spot on (:

To answer the remaining questions regarding worlds, and nits:
> High-level question: are the worlds in the "World Examples" section of messaging.wit.md just example worlds that you could write, but aren't being standardized, or are these worlds being standardized alongside the interfaces in the "Interfaces" section? If the former, maybe replace the "wasi:messaging/*" in their names with an unofficial-sounding kebab-name like my-world so it's a bit more clear. If the latter, maybe remove "Examples" from the section title.

> Thanks a lot for the explanation @Mossaka, that makes sense. The remaining detail I don't understand is why the pull interface is exported by the sub-pull world instead of being imported: for long-running instances, I would expect the export to be something like a main function that didn't take any event-specific parameters.

The first two worlds were there to illustrate that some interfaces (while separate) are intrinsically linked (i.e., `sub`/`basic/pull` and `sub`/`stream/pull`). Also, I should add that they are written separately in the first place because of the wide-range of use-cases for messaging – some providers only implement streaming pull, while others implement basic pull, and so forth. That said, in retrospect, I don't see much sense in doing an `export` of `pull`, so I'll fix that.

For the remaining worlds, I'll make it clearer that they are not being standardized and, instead, are just examples!

> To allow efficient batch-processing of events via the push interface, instead of having a separate interface with a stream function, I'd recommend instead adding a second streaming function to wasi:messaging/push that has a stream param type.

I mentioned why I avoided doing this in my previous answer, but, yeah~ For completion, that's hard to do because not every provider that has an `on-receive`, has a `stream-receive`.

> It might be nice to rename wasi:messaging/push/on-receive to wasi:messaging/handler/handle for symmetry with wasi:http/handler (they're both serving the same purpose, just different message types)?

> One world I'd expect to see standardized but isn't quite is:
world "wasi:messaging/service" {
  import pub: "wasi:messaging/pub"
  import sub: "wasi:messaging/sub"
  export message-handler: "wasi:messaging/handler"
}


I like that, I'll update it!

---

Regarding the removal of `pull`-based messaging, I'm okay w/ it, but I think there is most definitely a need for it and it should land somewhere at some point. This might be a good topic to bring up in the upcoming WASI Subgroup Video Meeting for a more in-depth discussion...

Issue #8, opened 2023-04-05
Current interface does not support a number of important use cases


autodidaddict:
I wasn't sure whether to put all of the use cases in their own issues or bundle it up. I can always split this one later if people want.
The core of this issue is that the current way in which wasi-messaging is designed (by current I mean what I can see in `main`, I'm not sure if there's work going on elsewhere) only supports a tiny fraction of the use cases required by application developers interacting with messaging systems.

## Do not mandate Cloud Events
Forcing the publisher of messages to only ever use cloud events is unnecessarily prohibitive. Only a small fraction of publications on message brokers are actually events, and an even smaller portion of that ever really need the shape of a cloud event. I understand the desire for an envelope, but I don't think it's the responsibility of this interface to enforce the envelope shape. When we get into requests (below), the mandate of a cloud event switches from inconvenient to a blocking impedance mismatch.

## Request/Reply
The current `consumer` interface only supports subscribe and unsubscribe. Consumers of message brokers need the ability to make a request and await a reply. This is perfectly normal to do synchronously so it wouldn't need to be done in collaboration with async/threads. I suggest supporting a call that looks like this:

```wit
request: func(b: broker, topic: string, data: list<u8>) -> result<list<u8>, error>
or...
request: func(b: broker, topic: string, timeout_ms: option<u8>, data: list<u8>) -> result<list<u8>, error>
```
As a convenience, we should probably give people the ability to make a request that expects multiple replies rather than just one. This is a basic necessity for message brokers that facilitates patterns like map/reduce, scatter/gather, polling, etc.
```wit
request-many: func(b: broker, topic: string, timeout_ms: option<u8>, max_results: option<u8>, data: list<u8>) -> result<list<list<u8>>, error>
```

The `channel` type doesn't feel appropriate here because that is an enum that distinguishes between queue and non-queue _subscriptions_. Making a request on a topic should never involve explicit knowledge if the target(s) are consuming via queue or not. In other words, as mentioned below, publication should be done on a topic, not a channel. The publisher/requester in a system should not have special knowledge of how that message will be consumed.

## Publish should be on a topic, not a channel
Depending on which message brokers people have used in the past, people come with preconceptions about what consumers and producers should know about in advance. As a general rule, a publisher or requester should have _no_ knowledge of the target of the message beyond that of the topic on which it is published. Whether there is a queue consumer, or multiple queue consumers, durable streams, or anything else listening on that topic should be _unknown_ to the client.

[...]

PR #9, opened 2023-04-06
`wasi-messaging` Refactor ft. stakeholder review

danbugs:
This PR contains changes across the all world interfaces and focuses to address feedback from:
- @stevelr ,
- @autodidaddict (#8),
- @clemensv , 
- @devigned , and
- @Mossaka .

Thank you everyone for your help and patience w/ this big refactor. 

Below, I'll highlight the major changes to this interface. To see it actually being used w/ a dummy implementation, please refer to: https://github.com/danbugs/wasi-messaging-demo .

## `types`

- `receiver`, and `sender` have been unified into `client`, and the `client`'s only function is to connect to some sort of named message exchange service.
- channel is no longer a variant w/ distinctions between broadcast and unicast channel – it's just a `type` alias to a string as the rest are consumer issues.
- added a `guest-configuration` `record` that contains a list of subscriptions that a specific guest owns, and an optional key-value pair list of 'extensions', which can contain specific config required for a particular implementor (e.g., partitions/offsets, etc..)
- `format-spec` is now an `enum` rather than a `string` `type` alias to ensure implementors exhaustively support/address different types of data envelopes.
- `message` no longer enforces CloudEvents. Instead, it's a `record` with a binary payload and `content-type`-type field (i.e., `format`) for deserialization. 

## `consumer`

- with the addition of `guest-configuration` and its' capability to allows us to set specific partitions/offsets to read from, `stream-consumer` and `basic-consumer` were unified into a single `consumer` interface.
- added `subscribe-try-receive`/`subscribe-receive` methods to fulfill the `request-reply` pattern – they aim to create an ephemeral subscription to receive from a channel until a specific timeout/indefinitely, respectively. Also, as for some implementors it's uncommon to return a single message, both methods return `list<message>`. 
- added `update-guest-configuration`, which allows for things like: setting a specific partition/offset to read messages from in the case of Kafka/EventHubs, or even just unsubscribe. It's meant to be a fit-all solution for any implementor-specific quirks to be addressed.
- added settlement methods for completing/abandoning messages – what the host should do when reaching these states could be configured thorugh `guest-configuration`

## `producer`

- just like for the consumer, `stream-producer` and `basic-producer` were unified into a single `producer`.
- to mirror `consumer`, `send` takes `list<message>`.

## `guest`

- as Wasm modules are not meant to be long-running instances, the guest interface was designed to be quickly called and thrown away. Here's the imagined workflow:
(1) the host calls `configure` to obtain a list of implementor-specific configurations (if needed), and a list of channels to subscribe and create a long-running client connection on the host side. At this point, the host can kill this component instance. 
(2) Once any messages come in for the subscriptions returned by `configure`,  the host can spawn up a new component and call the `guest`'s `handler` to take care of the incoming message. Equal to the motivations presented in `consumer` and `producer`, `handler` receives a `list<message>`.


wit/basic-consumer.wit
stevelr, line 1: are other types of consumer/producers planned besides `basic-consumer` and `basic-producer`? In other words, why not just `consumer` and `producer`? If other variations are planned, can they be documented as part of this proposal?

wit/basic-consumer.wit
```diff
@@ -0,0 +1,7 @@
+/// A basic consumer, commonly used for load-balancing.
+default interface basic-consumer {
+	use pkg.types.{message, error, receiver}
+
+	/// `receive-message` satisfied pull-based messaging requirements
+	receive-message(r: receiver) -> result<message, error>
```
stevelr: simplification suggestion:
   `receive` instead of `receive-message`

wit/basic-consumer.wit
```diff
@@ -0,0 +1,7 @@
...
+}
```
stevelr: Since so many things can go wrong in distributed messaging, it's useful to have a 
  `receive-timeout` method that accepts a timeout parameter,
so the program can avoid hangs when things do go wrong.

There's a subsequent problem of how to configure the timeout value, but I assume configuration is out of scope for this interface. Even if the timeout is hard-coded, it's better than not having a timeout option.

wit/basic-producer.wit
```diff
@@ -0,0 +1,6 @@
+default interface basic-producer { 
+    use pkg.types.{message, error, sender}
+
+    send-message: func(s: sender, m: message) -> result<_, error>
```
stevelr: Simplification suggestion: `send` and `send-many`

stevelr:
please bear with the the long-ish response. I want to articulate some goals, so if there's a difference in perspective we can address that before getting stuck in the details of apis.


The difference between basic- and stream- seems to be a difference in metadata associated with the message.
    - both include subject (aka entity), format (aka type), and optional metadata
    - `stream-` also includes partition and offset

Is it possible to combine and simplify so that we end up with only one type of producer and one type of consumer?

If we unwrap all the layers, the full set of metadata considered by the various send/receive apis here is
  - message data (`list<u8>`)
  - subject (`string`)
  - type, e.g., mime type (`string`)
  - partition (`u32`)
  - offset (`u32`)
  - optional metadata (`list<tuple<...>>`)

Sender and Receiver objects embody the subject name. I think that abstraction adds an unnecessary layer of complexity. Some messaging systems (at least NATS and MQTT, possibly others) permit subscribing to a set of topic with wildcards, so subject needs to be delivered with each message. In other words, in the general api, a subscriber can't receive  a message without also receiving its subject. Similarly, a single producer can often transmit messages on multiple subjects. In both cases, if subject were a string parameter of `send` and `receive`, it would simplify usage and remove the need for the `sender` and `receiver` objects.

NATS, PubSub, an MQTTv5 include optional attributes (aka headers or properties) with each message so an optional tuple of properties also seems warranted, for both `send()` and `receive()`

The format parameter is sometimes useful. A few are enumerated: MQTT, HTTP, etc. If a format parameter is part of the api, it needs a 'raw' variant as a catch-all if none of the other formats applies. Perhaps a mime type would be more inclusive and handle even more variants. To decide whether it should be included in the send/receive signatures (or a message struct),  and incurring the complexity and expense, we want to know how often it's needed. In most cases I've seen, the format is effectively communicated by the subject name and the documentation for the message stream. For example, if you're watching for changes to a database based on a set of keys, or watching a file system for changes to a file, the db and fs specify the format of the event. It's often specified along with the subject name of the event stream, in which case the publisher and subscriber already know the data format.

There are always tacit aspects of a message protocol that are part of its spec, in the documentation. If the implementers of the publisher and consumer have the same understanding of the protocol, and it doesn't very per message, it doesn't need to be sent with each message. Other protocol attributes that are often not explicit fields in the message are compression algorithm and encryption algorithm. In cases where additional or optional metadata is required to describe format, or other message attributes, the format can be included in a property map with other metadata (for example, the Content-Type value in the headers map in an HttpRequest), or the format and other metadata can be put into a wrapper object that becomes the new message (for example, a CloudEvents message has explicit metadata with a `data` field for the inner payload). If, in the vast majority of cases, format is tacit, or it can be included in metadata or the message object, it doesn't need its own field in the wasi messaging interface, either as a field of `message` or as a function parameter.
    
That leaves partition and offset. Although they are used with Kafka, and Kafka may be a common expected use case, it's not needed for a lot of other implementations, and harder to justify for inclusion in a common api used by all implementations. Every user of this api would have to go through a decision process on whether to interpret the offset parameter as a sequence, and wonder if they need to implement message reordering. If our goal is to make it easy to create lightweight WASM components, we don't want users to even have to think about message reordering - it should be in a service layer.

How can we enable WASM developers to use Kafka partitions, and QOS, and TTL, and other variations of messaging services that might be critical to their needs? In the short term, we have ways to handle variations: on the producer side, create components that accept additional service-specific parameters and pass a wrapper object as the message to the basic interface. On the consumer side, use a component that decodes the wrapper object and calls a subscriber that knows how to handle the extra parameters for receive. Similarly, wrappers can be created when subscribing, or when obtaining a broker handle. 

**I'm hopeful that the initial set of wasi interfaces can have a high enough level of abstraction to be broadly useful and simple to use, where api designers apply a high bar before deciding to add parameters that increase complexity for all users. Ideally, keeping these shared wasi interfaces simple will make them broadly applicable, and help adoption by shorten the learning curve for all developers.**

One way to future-proof the basic api would be to allow config parameters when opening a broker. That could be an escape hatch used by wrapper components that need to pass implementation-specific parameters into the basic interface.

To summarize, I propose:
 - combining basic-producer, basic-consumer , stream-producer, stream-consumer  into `producer`, `consumer`
 - combining sender, receiver into string, and including subject in each invocation of `send` and `receive` (as a function parameter or field of the `message` object)
 - removing `partition`, `offset`

This leave us with something like this: 
```
    interface broker{
        open(config: option<metadata>) -> result<broker, error>
        publish() -> result<producer, error>
        /// subscribe to a subject or wildcard set of subjects
        subscribe(subject: string) -> result<consumer, error>
    }
    interface producer {
        use self.types.{error, message}
        send(message: message) -> result<_, error>
    }
    interface consumer {
        use self.types.{error, message}
        receive() -> result<message, error>
        receive_timeout(timeout: timeout-ms) -> result<message, error>
    }
    interface types {
        record message {
            subject: string,
            data: list<u8>,
            metadata: option<metadata>,
        }
        
        type metadata = list<tuple<string, string>>
        /// timeout in milliseconds
        type timeout-ms = u32
        
        variant error {
            timeout,
            ...
        }
    }
```

wit/basic-consumer.wit
Mossaka: Strongly agree that we should have a receive-timeout method

wit/types.wit
```diff
@@ -1,41 +1,47 @@
...
+    /// An entity may be a queue or a topic and is identifiable by name.
+    type entity = string
+
+    /// A message contains binary data, a format specification, and a metadata decoration of key/value pairs.
+    ///     - The format could be: "CloudEvents/1.0", "HTTP/1.1" / "HTTP/2" / "HTTP/3", "AMQP/1.0", "MQTT/3.1.1" / "MQTT/5.0", "Kafka/0.11"
+    ///     as defined per: https://github.com/clemensv/spec/blob/registry-extensions/registry/spec.md#message-formats
+    record message {
+        data: list<u8>,
+        format: string,
+        metadata: option<list<tuple<string, string>>
```
clemensv: Good to allow duplicates for the metadata keys. The keys should allow for dot-notation for qualifying properties such that you can address, for instance, "properties.reply-to" in AMQP directly. 

Unqualified properties should have a default property-bag depending on protocol, which would then commonly map to whatever "user defined properties" are, e.g. "application-properties" in AMQP or "user property" in MQTT.

wit/types.wit
```diff
@@ -1,41 +1,47 @@
...
-        queue(string),
-        topic(string)
+    /// An entity may be a queue or a topic and is identifiable by name.
+    type entity = string
+
+    /// A message contains binary data, a format specification, and a metadata decoration of key/value pairs.
+    ///     - The format could be: "CloudEvents/1.0", "HTTP/1.1" / "HTTP/2" / "HTTP/3", "AMQP/1.0", "MQTT/3.1.1" / "MQTT/5.0", "Kafka/0.11"
+    ///     as defined per: https://github.com/clemensv/spec/blob/registry-extensions/registry/spec.md#message-formats
+    record message {
+        data: list<u8>,
```
clemensv, line 40: Since `data` needs to be qualified for any consumer to understand it, `content-type` must be special-cased as a property right on this interface. You need to be able to decode `data` without digging into the property bag. We made the same call for CloudEvents.

wit/types.wit
```diff
@@ -1,41 +1,47 @@
...
-    /// - queue: competitive consumers, and
-    /// - topic: non-competitive consumers.
-    variant channel {
-        queue(string),
-        topic(string)
+    /// An entity may be a queue or a topic and is identifiable by name.
+    type entity = string
+
+    /// A message contains binary data, a format specification, and a metadata decoration of key/value pairs.
+    ///     - The format could be: "CloudEvents/1.0", "HTTP/1.1" / "HTTP/2" / "HTTP/3", "AMQP/1.0", "MQTT/3.1.1" / "MQTT/5.0", "Kafka/0.11"
```
clemensv: I like that call

wit/basic-consumer.wit
clemensv: We need settlement (complete, abandon, deadletter or accept, release, reject) as first-class constructs for all queue-based interactions. The decision on how to settle a message is often dependent on the application outcome.

In queues, all messages are settled singly. MQTT also requires accept/reject (positive/negative ACK) for QoS 1/2.

You could put a `receive-context` on the message that has appropriate settlement methods. That would then also work with the reactive API style below.

wit/basic-producer.wit
clemensv: I would use `send-batch` for the method that takes a list. `send-many` says "multiple sends in a loop", but what you'll want the provider to do is to do a batched transfer.

wit/basic-consumer.wit
clemensv: All methods that interact with the network need a timeout option.

wit/handler.wit
```diff
@@ -1,7 +1,9 @@
...
-    use pkg.types.{event, error}
+    use pkg.types.{message, error}
 
-    /// Creates an on-receive handler push-based message delivery.
-    on-receive: func(e: event) -> result<_, error>
+    /// Configures the handler, and must return a receiver obtained through `open-receiver` function implemented by the host.
+    configure-handler: func() -> result<receiver, error>
+
+    /// Creates an on-receive handler for push-based message delivery that may receive a batch of messages called by the host when messages are sent.
+    on-receive: func(m: list<message>) -> result<_, error>
```
clemensv: See above regarding passing context.

wit/messaging.wit
```diff
@@ -1,5 +1,11 @@
-default world messaging {
-	import producer: pkg.producer
-	import consumer: pkg.consumer
+default world load-balancer {
```
clemensv: 🤨 "load-balancer" is an odd term for this.

wit/stream-consumer.wit
```diff
@@ -0,0 +1,7 @@
+/// A stream consumer, commonly used for event streaming.
+default interface stream-consumer {
+	use pkg.types.{message, error, receiver}
+
+	/// `receive-message` satisfied pull-based messaging requirements
+	read-message(r: receiver, partition: u32, offset: u32) -> result<message, error>
+}
```
clemensv: the consumer needs to be set up with a configuration that specifies things like partitions and offsets and subsequently  just read. 

part of that setup needs to be a configuration of a) a checkpoint abstraction and b) a lease abstraction for whether the client still owns the partition. 

I expect future event streams to kill partitions altogether as a user-facing concept. You should also anticipate "offset" to become a query. "u32" is already false for EH and Kafka, which allow for start,end,time.

This needs work.

wit/stream-producer.wit
```diff
@@ -0,0 +1,6 @@
+default interface producer { 
+    use pkg.types.{message, error, sender}
+
+    send-message: func(s: sender, m: message, partition: u32, offset: u32) -> result<_, error>
+    send-message-batch: func(s: sender, m: list<message>, partition: u32, offset: u32) -> result<_, error>
```
clemensv: The producer interface should be the same for queues and streams. stash the partition info into the factory. you also don't send to offsets. that is a consumer thing.

clemensv (commented):
It is very hard to consume this as a set of interface definitions. 

Can we turn this into a doc that is a list of example snippets for various scenarios? The interfaces should then trivially follow from the snippets.

wit/basic-consumer.wit
devigned: If a timeout is going to be added, would it make sense to have a structure closer to the Golang `context.Context`?

See: https://pkg.go.dev/context

Context would be useful for not just timeouts, but also for telemetry, trace propagation.

wit/types.wit
danbugs: ~closing as outdated.

wit/consumer.wit
```diff
@@ -1,10 +1,26 @@
...
+    /// Blocking receive until message with ephemeral subscription
+    subscribe-receive: func(c: client, ch: channel) -> result<list<message>, error>
+
+    /// 'Fit-all' type function for updating a guest's configuration – this could be useful for:
+    ///     - unsubscribing from a channel,
+    ///     - checkpointing,
+    ///     - etc..
+    update-guest-configuration: func(gc: guest-configuration) -> result<_, error>
+
+    /// A message can exist under several statuses:
```
autodidaddict, line 17: This feels like an inconsistency in the abstraction level. When I think about the goals of these interfaces, one of them is typically to hide certain implementation details from people developing the components. 

What worries me about having `complete` and `abandon` is that it is surfacing an ack/nack pattern that can be problematic. Firstly, not all brokers implementing this interface may support acks and nacks at all. Secondly, some brokers may support ack/nack only under certain conditions and in certain configurations, creating unpredictability in how an application may perform from one provider to another or even from one environment to another.

Item 2 in this list even uses the word `queue`, which may or may not be how the implementer of this interface manages messages.

Put more simply: as a developer using wasi-messaging, I don't want to care about acks or nacks. I want to publish messages, make requests, and receive messages from subscriptions.

wit/types.wit
```diff
@@ -1,41 +1,44 @@
...
-    /// - queue: competitive consumers, and
-    /// - topic: non-competitive consumers.
-    variant channel {
-        queue(string),
-        topic(string)
+  
+    /// Format specification for messages 
+    ///  - more info: https://github.com/clemensv/spec/blob/registry-extensions/registry/spec.md#message-formats
+    ///  - message metadata can further decorate w/ things like format version, and so on.
+    enum format-spec {
```
autodidaddict, line 29: As mentioned in issue #8 , I don't think the payload type should be surfaced as part of the interface. If we do this, then we'll have to change the interface every time we find a new payload type that we want to support or, more likely, everyone will just use `raw`.

I don't expect my brokers to care what shape my bytes take internally, I only expect them to deliver my bytes according to my needs.

autodidaddict (changes requested):

autodidaddict (changes requested):
See individual comments for specifics.

subscribe-receive and subscribe-try-receive aren't named using common idioms from message broker patterns. A subscribe-receive might be more self-documenting if it was named `request`. Also, if both of these functions return a list of messages, then there's no way to reliably implement the request-(single)response pattern, at least from looking at the comments.

Choosing whether you're going to await multiple responses to a single request, or whether you're "done" waiting after you receive the first response has a big impact on application behavior. If people can't easily get a single reply, then they're going to write wrappers that do this.

I'd still like to see `request` (subscribe-receive, but limited to 1 response) and `request-multi` (subscribe-receive, but you can specify an upper bound or wait for the timeout). After simple publication, I'd wager that these two are the most common uses for message brokers.

wit/types.wit
autodidaddict: Another use case that this format spec can interfere with is using different transports. Let's say I have a component and I want to run it both in a cloud and on a constrained edge device. In the cloud, this component can use a robust messaging transport. On the constrained device, it probably uses MQTT. The component should not need to know if it's running in the cloud or a constrained device, it should simply execute its business logic while the use of MQTT or Kafka or anything else is an invisible detail. If we have to specifically embed the format type into our compiled code, it blocks the benefits of portability that WebAssembly is so famous for.

autodidaddict:
I agree with @clemensv 's comment from last month that showing some sample code might help make writing the wit easier. I'd like to be able to do the following in my Rust wit-bindgen code:

```rust
let reply = messaging::request("service.api", my_bytes, Some(2_000))?; // wait for a reply for max of 2s

// collect pong replies, stopping after 2s or 30 replies, whichever comes first
let pong_collection = messaging::request_multi("ping", vec![], 30, Some(2_000))?; 
```
If I want to use an envelope structure here, then it's my responsibility as the component developer to do that. The broker should remain unaware of the shape of the innards of my messages, and doesn't need to know how to decode my envelopes.

autodidaddict:
@clemensv in #8 's comment section you mention that the pattern of request/reply as well as things like scatter-gather are application-specific patterns and shouldn't pollute the foundation. Would you then also assert that the latest revision's `complete` and `abandon` pair are to much like an app-level ack/nack pair? It feels like if request/reply, scatter/gather are app patterns, then so too is ack/nack. 

One of the litmus tests I've been using around these `wit` interfaces is that if it looks like at least 80% of the consumers  of the interface would likely have to write a wrapper to add a super-common pattern, we should consider including that pattern in the interface. This means the difference between being able to use wit-bindgen code off the shelf versus having to write a wrapper library.

Thoughts?

wit/types.wit
danbugs: What are your thoughts on simply making `format-spec` a `string` alias then instead of an enum?~ similar to `content-type` in HTTP headers.

danbugs:
About request/reply – While I understand that it is only one pattern of many, I do believe it's a very commonly used pattern (afaik, more than scatter/gather), and, so, it might align with the goal of `wasi-cloud-core` interfaces of covering 80% of user applications that use x capability.

I'll work on adding some example code to this PR for ease of review next!

autodidaddict:
Yeah, I lean toward request/reply being more fundamentally required than the other patterns (like gather).

wit/types.wit
autodidaddict: Why don't we just call it `content-type` and make it a string, this way it's easier for people to make that analogy?

clemensv:
> @clemensv in #8 's comment section you mention that the pattern of request/reply as well as things like scatter-gather are application-specific patterns and shouldn't pollute the foundation. Would you then also assert that the latest revision's `complete` and `abandon` pair are to much like an app-level ack/nack pair? It feels like if request/reply, scatter/gather are app patterns, then so too is ack/nack.
> 
> Thoughts?

Settlement is a core function of all messaging systems even if it comes in slightly different shapes. And it's a direct interaction with the broker and/or peer endpoint. 

Request/reply is an application-level contract that is projected through the infrastructure and not even AMQP or MQTT5, which have correlation metadata for reply paths, have firm opinions about how those are to be used at the protocol level and punt that up to the app as a problem. MQTT5 has a [non-normative section](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901252) about it, AMQP is completely silent beyond describing the properties.

autodidaddict:
> Settlement is a core function of all messaging systems even if it comes in slightly different shapes. And it's a direct interaction with the broker and/or peer endpoint.
> 

Do MQTT and websockets support settlement? 

I suppose there's no harm in an implementor of the interface doing a no-op in response to the complete/abandon requests. 

If the interface doesn't support explicit request/reply semantics, then what provision is there in the interface to provide hints/metadata as to the reply mechanism? For example, most brokers that support request/reply intrinsically require you to supply the topic on which the reply is to be published.

Please note that I'm not contradicting opinions, I'm just trying to tease out multiple perspectives based on people's experiences with multiple different brokers. It's easy for me to assume "all brokers" work a certain way if I've only worked with NATS, Kafka, AWS's SQS, and RabbitMQ. I'm genuinely surprised by the notion that request/reply isn't a fundamental concept supported by all message brokers.

clemensv:
Yes, MQTT does support settlement in the form of QoS 1 and Qos 2 handshakes. 

WebSockets is a stream protocol with framing layered on top of HTTP and generally requires an overlaid application protocol to be used for messaging. Putting JSON into text frames is where an overlaid application protocol starts, and if you add a "got it!" reply, you've got settlement. Frameworks like socket.io or SignalR (there are many) have settlement as feature ([example](https://socket.io/docs/v4/#acknowledgements))

AMQP and MQTT can be layered on top of WebSockets as application protocol layer and that is quite common.

Regarding request/reply: Some of brokers you mention support request/reply as a construct over their underlying primitives. Kafka doesn't. NATS uses reply subjects and [Rabbit uses reply queues](https://www.rabbitmq.com/tutorials/tutorial-six-python.html), but also has a way to create [reply queues on the fly](https://www.rabbitmq.com/direct-reply-to.html). You can construct something similar with JMS.

However, the core messaging APIs generally remain async and one-way. Even for the "Direct Reply-To" feature of Rabbit, the replies are pulled from a pseudo-queue. 

You would have a similar model for AMQP 1.0 in brokerless peer-to-peer mode that allows you to create bi-directional communication connection but with independent flows over opposing links. Requests and replies would be matched by the overlaid client. The messaging path is oblivious to the app-level correlation by design.

wit/types.wit
clemensv: For an API abstraction, you will want to have access to the features of the broker you want to deal with. For a common event format like CloudEvents, you want to abstract as much of that away. If you want a common envelope for use with an API, pick a common envelope format like CNCF CloudEvents. We* did a lot of work to find commonalities and figure out transport bindings and mappings. 

A unified API that works across products should have a foundational message that is even more fundamental than CloudEvents (payload content-type, payload body, metadata map and map of maps) so that anyone can understand the payload without having to understand envelopes or protocols at all. For most apps, payload is the most important thing.

Yet, the unified API should not aim to hide the details of the underlying brokers and their specific metadata because if it does, developers will quickly jump back to product-specific APIs. The experience needs to provide a balance of unified experience and maximum product/feature access fidelity at the same time. If you use Kafka, you need to be able to set the record/partition key and if you use an AMQP broker you need to be able to set the group-id. 

Therefore, having a selector that tells you want to expect in the metadata section, similar to content-type helping with what to expect in the payload section, does make sense. 

* (I am one of the architects of CNCF CloudEvents)

autodidaddict:
Thanks for filling in the blanks @clemensv . Some aspects of this are much clearer now.

danbugs:
@autodidaddict , and @clemensv ~ As requested, I've added some examples.

Also, @Mossaka , I've updated the interface to the latest WIT version for the examples, so I'll be closing #12 .

examples.md
```diff
@@ -0,0 +1,178 @@
+# Examples
+
+## Example 1: Guest Skeleton Usage
```
clemensv, line 3: I've been doing messaging in all facets for 30+ years and can't tell what a "guest" is. Is that a WebAssembly specific term?

examples.md
```diff
@@ -0,0 +1,178 @@
...
+struct MyGuest;
+
+impl guest::Guest for MyGuest {
+    fn configure() -> Result<GuestConfigurationResult, u32> {
+        // This function will be called by the host, who will be maintaining a 
+        // long-lived client connection to a broker or other messaging system.
+        // The client will be subscribed to channels a, b, and c) with no extra configuration.
+        // As soon as configuration is set, the host should kill the Wasm instance.
+        Ok(GuestConfigurationResult {
+            channels: vec!["a".to_string(), "b".to_string(), "c".to_string()],
```
clemensv, line 97: What does "subscribe to channel a" entail here? What criterion are messages separated or dispatched by? 

This construct could mean three things:
a) there are three distinct entities named a, b, c in the broker that this connection attaches to
b) there are three distinct sessions/partitions a, b, c in the broker on a single entity that this connection attaches to
c) there are three types of messages forming logical channels (multiplexed) that come over the connection

Which is these is it?


You'll need to be able to cover all three, IMO

examples.md
```diff
@@ -0,0 +1,178 @@
...
+            ..Default::default()
+        })
+    }
+
+    fn handler(ms: Vec<MessageResult>) -> Result<(), u32> {
+        // Whenever a message is received on a subscribed channel (from configure()),
+        // the host will call this function. Once the message has been handled,
+        // the host is expected to kill the Wasm instance.
+        
+        for m in ms {
```
clemensv, line 107: For examples that try to explain an API, it's very helpful to be far less frugal with letters in variable names.

examples.md
```diff
@@ -0,0 +1,178 @@
...
+
+            // match on message metadata for channel name
+            match m.metadata {
+                Some(metadata) => {
+                    for (k, v) in metadata {
+                        if k == "channel" {
```
clemensv, line 113: There's an assumption here that "channel" is in the metadata. Is that assumed to map from/to a construct in the messaging infrastructure or is that end-to-end application data? if so, how does "channel" interact with partitions or subscriptions in the broker? For instance, with MQTT or NATS, how do you specify the subject/topic to the broker for selecting a,b,c into this channel?

examples.md
```diff
@@ -0,0 +1,178 @@
...
+                                    // abandon message
+                                    consumer::abandon_message(m).unwrap();
+                                }
+                                "b" => {
+                                    // handle message from channel b
+                                    // [...]
+
+                                    // request-reply from channel d
+                                    let client = connect("some-broker").unwrap();
+                                    let msgs = subscribe_try_receive(client, "d", 100).unwrap();
```
clemensv, line 134: You have a "subscribe" gesture here but you have a "change configuration" gesture a bit further up that also seems to be modifying what you subscribe to, so I'm confused.

there are several concepts that are quite universal in messaging across brokers and APIs that you need clear mappings for:

a) Config/Options/Credentials -- a set of information required to locate and connect with the broker and supply upfront options such as prefetch buffers and the like
b) Connection - a representation of the network connection including, if provided by the abstracted API or implementation, its connection state
c) Session - a representation of a session or link layered over the network connection. Protocols that have the ability to reconnect (MQTT and AMQP for instance) have such abstractions that allow maintaining a relationship between the parties in spite of the connection dropping out underneath. You also have that for HTTP as a key concept.
d) Link/Subscription/Sender/Receiver - a representation of a particular communication route to an entity, usually either for sending or receiving, sometimes coupled with a volatile subscription that is established with the link to that entity. Sometimes a "subscription" you attach to is an already existing entity in the broker (durable subscription, consumer group/partition). Mind that a client may not be authorized to create such a subscription.

examples.md
```diff
@@ -0,0 +1,178 @@
...
+
+                                    // complete message
+                                    consumer::complete_message(m).unwrap();
+                                }
+                                "c" => {
+                                    // handle message from channel c
+                                    // [...]
+
+                                    // send message to channel d
+                                    let client = connect("some-broker").unwrap();
```
clemensv, line 150: Examples should echo best practice. "connect" to a different broker while you are handling a message is never and nowhere a good idea. 

Reason why I point that out is that best practice ought to shape the API. Messaging connections are generally longer lived since their protocols are sessionful.

clemensv (commented):
The examples and the API need to be better informed by prior art and it needs to become apparent why this approach of a new API is needed and superior. 

Take existing examples of client interactions (especially receiving, since sending is mostly all the same everywhere) with Kafka, Rabbit, Service Bus, SQS, Mosquitto, Event Grid, and NATS using their native SDKs and juxtapose those with the interfaces developed here and explain how the respective concepts match.

examples.md
autodidaddict: Yes. Guest refers to the WebAssembly module while host refers to the execution runtime, e.g. wasmtime or something built on top of that.

guest271314:
I think this can be as simple as Native Messaging protocol https://developer.chrome.com/docs/extensions/mv3/nativeMessaging/#native-messaging-host-protocol.

> The same format is used to send messages in both directions; each message is serialized using JSON, UTF-8 encoded and is preceded with 32-bit message length in native byte order.

We don't need subject or metadata as separate fields. Just a `Uint32Array` containing message length followed by the message as JSON. That's it.

autodidaddict:
> We don't need subject or metadata as separate fields. Just a Uint32Array containing message length followed by the message as JSON. That's it.

This may be too simple of a protocol. Components need to be able to dictate the topic on which messages are published. We can't just assume that there's one catch-all channel like `stdio`. Without being able to indicate a publication subject/topic, a huge swath of common scenarios and use cases won't be possible for components.

guest271314:
> This may be too simple of a protocol.

Simple and effective without being over-specified or over-engineered needlessly.

We can stream real-time audio using the protocol, where you and users will notice gaps and glitches in playback which can be hidden in video. 

Here's the same algorithm in C, C++, JavaScript (QuickJS), and Python where it  is unobservable client side which programming language(s) are being used as the host. 

- https://github.com/guest271314/captureSystemAudio/blob/master/native_messaging/capture_system_audio/capture_system_audio.c
- https://github.com/guest271314/captureSystemAudio/blob/master/native_messaging/capture_system_audio/capture_system_audio.cpp
- https://github.com/guest271314/captureSystemAudio/blob/master/native_messaging/capture_system_audio/capture_system_audio.js
- https://github.com/guest271314/captureSystemAudio/blob/master/native_messaging/capture_system_audio/capture_system_audio.py

and some other Native Messaging hosts https://github.com/guest271314/NativeMessagingHosts.

> Components need to be able to dictate the topic on which messages are published. 

Then do that. 

Whether using a specific name for a host, e.g., `nm_c_wasm` that does one specific task, echo back 1 MB that the client sent to the host https://github.com/guest271314/native-messaging-webassembly/blob/main/nm_c_wasm.json, or open-ended, e.g., the host does whatever you tell it to do

```
const port = chrome.runtime.connectNative('capture_system_audio');
port.onMessage.addListener((message) => {
  // Do stuff with message, e.g., [0, 255]
});
port.onDisconnect((port) => {
  if (port.error) {
    // Handle port disconnecting or error
  }
});
port.postMessage('parec -d @DEFAULT_MONITOR@');
```

and

```
let text = `So we need people to have weird new
ideas ... we need more ideas to break it
and make it better ...

Use it. Break it. File bugs. Request features.

- Real time front-end alchemy, or: capturing, playing,
  altering and encoding video and audio streams, without
  servers or plugins!
  by Soledad Penadés
   
von Braun believed in testing. I cannot
emphasize that term enough – test, test,
test. Test to the point it breaks.

- Ed Buckbee, NASA Public Affairs Officer, Chasing the Moon

Now watch. ..., this how science works.
One researcher comes up with a result.
And that is not the truth. No, no.
A scientific emergent truth is not the
result of one experiment. What has to
happen is somebody else has to verify
it. Preferably a competitor. Preferably
someone who doesn't want you to be correct.

- Neil deGrasse Tyson, May 3, 2017 at 92nd Street Y

It’s like they say - if the system fails you, you create your own system.

- Michael K. Williams, Black Market

1. If a (logical or axiomatic formal) system is consistent, it cannot be complete.
2. The consistency of axioms cannot be proved within their own system.

- Kurt Gödel, Incompleness Theorem, On Formally Undecidable Propositions of Principia Mathematica and Related Systems`;

const port = chrome.runtime.connectNative('capture_system_audio');
port.onMessage.addListener((message) => {
  // Do stuff with message, e.g., [0, 255]
});
port.onDisconnect((port) => {
  if (port.error) {
    // Handle port disconnecting or error
  }
});
port.postMessage({cmd:`espeak-ng -m -v Storm --stdout`, input:`"${text}"`};);
```

And/or when networking is used simply use static URL endpoints, or query string parameters, or use QUERY. Though HTTP(S), QUIC, etc. have all kinds of extra headers and extraneous data to parse before we get to the data.

Using the Native Messagin protocol as long as the producer is expecting whatever the consumer throws it it, we can simply call the approriate commands to stream to `stdout`, or pass commands or even a socket connect on to the target endpoint(s).

Or, as demonstrated above, pass whatever you want, e.g., arbitrary shell commands as a string or JSON (JavaScript Object Notation), or encoded text or other data as `Uint8Array` or other `TypedArray` serialized to the JSON format `[0, 65536]`.

I am relatively certain you have not come up with use-cases the Native Messaging protocol cannot produce to and consume from efficiently.

guest271314:
>  We can't just assume that there's one catch-all channel like `stdio`. 

Ultimately all programs write to stdio or a file, or stream or do something the user can observe.

Sure we can. It's up to the host to handle the message, or not. This is how I check if a command passed to a Deno Native Messaging host should be executed or not - in the local HTTPS server the host starts https://github.com/guest271314/native-messaging-espeak-ng/blob/deno-server/local_server.js. 

```
const commands = new Set();
commands.add('espeak-ng');
```

```
const text = await request.json();
      const json = text.cmd.split(' ').concat(text.input);
      const program = json.shift();
      if (commands.has(program)) {
        const command = new Deno.Command(program, {
          args: json,
          stdout: 'piped',
          stdin: 'piped',
        });
        const process = command.spawn();
        const reader = process.stdout.getReader();
        body = new ReadableStream({
          async pull(c) {
            const { value, done } = await reader.read();
            if (done) {
              c.close();
            }
            c.enqueue(value);
          },
          async cancel(reason) {
            process.kill('SIGKILL');
            server = null;
            controller = null;
          },
        });
      }
      return new Response(body, responseInit);
```

Now I created this server to test and compare Fetch (Standard) API and Streams (Standard) API to Native Messaging locally.  The client can't tell the difference. So why use a more verbose and complex communication system with multiple layers of metadata (headers, etc.)?

There are far less moving parts using the Native Messaging host only. Technically we can establish a TCP socket and use that as the communication mechanism, similar to `WebSocketStream`, less moving parts than Fetch; both can be bi-directional streams. Less moving parts still in a Native Messaging hosts that just does what you tell it to do, whatever that might be, and streams  to stdout https://github.com/guest271314/native-messaging-deno/blob/main/nm_deno.js.

examples.md
danbugs: This will depend on the implementation. 

~extra required information can be passed via `extensions`, and it will be the implementor's job to specify if they required any `extension` parameter to set these subscriptions.

examples.md
danbugs: Yes – For the purpose of this example, the implementor is adding a `channel` key to the message's metadata.

How the channel interacts w/ partitions is completely up to the implementor, and I'm hoping the interface can remain generic enough to display that. This is in particular to cover implementors that don't concern themselves with partitions.

`channel` is also a generic name that can represent a queue or a subscription.

examples.md
danbugs: > You have a "subscribe" gesture here but you have a "change configuration" gesture a bit further up that also seems to be modifying what you subscribe to, so I'm confused.

`subscribe_try_receive` sets an ephemeral subscription to receive from a channel for a specified amount of time. `update_guest_configuration` can add new subscriptions or update anything that was set in the guest's `configure`.

About the rest the rest of your comment, like mentioned in the `wit/consumer.wit`, `update_guest_configuration` is a:
```
    /// 'Fit-all' type function for updating a guest's configuration – this could be useful for:
    ///     - unsubscribing from a channel,
    ///     - checkpointing,
    ///     - etc..
```

I hadn't given sessions much thought, but I did intend it to cover points (a), (b), and (d). This could be temporary, and a first step to breaking it apart onto other functions down the road.

danbugs:
@guest271314 ~ I think it might be helpful for you to open a separate issue highlighting how the interface should look like in accordance to `NativeMessaging` and how other implementors (e.g., Azure Service Bus, NATS, Kafka, Mosquitto, etc.) would utilize it to perform required actions highlighted in this interface with the simpler foundation.

danbugs:
@autodidaddict , and @clemensv ~ As this PR has already been open for a while, I'm debating concluding it as is, and creating separate issues to allow for more targeted PRs, thoughts? I'd really appreciate your help with creating issues that you think are pertinent. Please, give me an OK, or let me know if there's anything you think should be concluded in this very PR.

autodidaddict:
> @autodidaddict , and @clemensv ~ As this PR has already been open for a while, I'm debating concluding it as is, and creating separate issues to allow for more targeted PRs, thoughts? I'd really appreciate your help with creating issues that you think are pertinent. Please, give me an OK, or let me know if there's anything you think should be concluded in this very PR.

I think merging and moving forward is probably a good idea. From what I can tell of the "current" files, the conflict between manually subscribing to a channel and returning a list of subscriptions from the configure function has been reconciled (I think we only have `configure` now). As long as it's internally consistent we can iterate via more issues and PRs.

I'm hoping to be able to build a prototype from this to get some "hands on" time that will potentially lead to future issues

danbugs:
I got approval from all current reviewers, so I'll be merging this PR. Issues will spawn soon to target anything that wasn't resolved yet.

PR #10, opened 2023-04-07
Include local use cases; remove redundant "service"

### LINKED ISSUE #4: Use case: Local usage
This will be useful to provide a means to create a message channel between local applications and browsers, e.g., Native Messaging https://github.com/guest271314/native-messaging-webassembly, not just "service providers, and or cloud providers)".

This can work something like `WebTransport` and `WebSocketStream` where we write to the application with `writable.writer.write()` and read with `readable.reader.read()`.

guest271314:
Fixes https://github.com/WebAssembly/wasi-messaging/issues/4.

Include local use cases; remove redundant use of the term "service".

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,11 +28,18 @@ The messaging interfaces aim to provide a generic and flexible way for producers
 
 ### Goals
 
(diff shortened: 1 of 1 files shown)


danbugs (approved):
LGTM

PR #21, opened 2024-02-26
feat(*): Updates messaging to support request-reply

thomastaylor312:
This makes several updates to the messaging interface. Initially the README said that this wasn't going to support request/reply, but based on my reading of the Kafka, NATS, MQTT, and SQS APIs, this is a fairly common pattern. Another piece of evidence here is what I've seen as a wasmCloud maintainer from our users. Request/reply is one of the more common things we see with a messaging service. Please note that this doesn't _require_ the use of a reply-to topic, just exposes it for use.

I also did a few other changes here. First is that I added the topic to the message. This was common across all systems and is often used by code to select the appropriate logic to perform. I also removed the format field as this didn't seem to be a common parameter across various services. We could definitely add a content-type member to this record in the future if needed, but I think much of that can be passed via the metadata field.

There are other things I might suggest some changes to, but I want to think on them some more and open some issues to discuss them first

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Overall, the messaging interfaces aim to make it easier to build complex and sca
 ### Non-goals
 
 - The messaging service interfaces do not aim to provide advanced features of message brokers, such as broker clustering, message persistence, or guaranteed message delivery. These are implementation-specific details that are not addressed by the interfaces.
-- The messaging service interfaces do not aim to provide support for every possible messaging pattern or use case. Instead, they focus on the common use cases of pub-sub and push-based message delivery. Other messaging patterns, such as request-reply or publish-confirm-subscribe, are outside the scope of the interfaces.
+- The messaging service interfaces do not aim to provide support for every possible messaging pattern or use case. Instead, they focus on the common use cases of pub-sub, push-based message delivery, and request-reply. Other messaging patterns, such as publish-confirm-subscribe, are outside the scope of the interfaces.
 - The messaging service interfaces do not aim to provide a specific implementation of a message broker. Instead, they provide a standard way to interact with any message broker that supports the interfaces.
 
 ### API walk-through
--- a/wit/types.wit
+++ b/wit/types.wit
@@ -23,22 +23,15 @@ interface messaging-types {
         extensions: option<list<tuple<string, string>>>
     }
   
-    /// Format specification for messages 
-    ///  - more info: https://github.com/clemensv/spec/blob/registry-extensions/registry/spec.md#message-formats
-    ///  - message metadata can further decorate w/ things like format version, and so on.
-    enum format-spec {
-        cloudevents,
-        http,
-        amqp,
-        mqtt,
-        kafka,
-        raw
-    }
-  
-    /// A message with a binary payload, a format specification, and decorative metadata.
+    /// A message with a binary payload and additional information
     record message {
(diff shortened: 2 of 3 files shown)


thomastaylor312:
Also, I still have no idea what tool it is we use that updates all of the autogenerated HTML docs, so if someone could direct me with that, it would be great!

wit/types.wit
```diff
@@ -23,22 +23,15 @@ interface messaging-types {
         extensions: option<list<tuple<string, string>>>
     }
   
-    /// Format specification for messages 
-    ///  - more info: https://github.com/clemensv/spec/blob/registry-extensions/registry/spec.md#message-formats
-    ///  - message metadata can further decorate w/ things like format version, and so on.
-    enum format-spec {
```
danbugs, line 29: `format-spec` wasn't part of the original interface, but was added during our stakeholder review. Here's some relevant comments you may want to take a look at:
(1) https://github.com/WebAssembly/wasi-messaging/pull/9#discussion_r1178045058
(2) https://github.com/WebAssembly/wasi-messaging/pull/9#discussion_r1206702667

README.md
```diff
@@ -45,7 +45,7 @@ Overall, the messaging interfaces aim to make it easier to build complex and sca
 ### Non-goals
 
 - The messaging service interfaces do not aim to provide advanced features of message brokers, such as broker clustering, message persistence, or guaranteed message delivery. These are implementation-specific details that are not addressed by the interfaces.
-- The messaging service interfaces do not aim to provide support for every possible messaging pattern or use case. Instead, they focus on the common use cases of pub-sub and push-based message delivery. Other messaging patterns, such as request-reply or publish-confirm-subscribe, are outside the scope of the interfaces.
+- The messaging service interfaces do not aim to provide support for every possible messaging pattern or use case. Instead, they focus on the common use cases of pub-sub, push-based message delivery, and request-reply. Other messaging patterns, such as publish-confirm-subscribe, are outside the scope of the interfaces.
```
danbugs, line 48: Adding support for `request-reply` was rejected during our stakeholder review - here's a relevant comment explaining why: https://github.com/WebAssembly/wasi-messaging/issues/8#issuecomment-1570286293

danbugs (commented):
As pointed out by the comments below, some of these suggestions were previously rejected by our stakeholder review. 

That said, I wouldn't mind re-evaluating depending on how folks feel - I've always been more of the opinion that our interfaces should better reflect usability.

wit/types.wit
thomastaylor312: Hmmm...I think agree with that second comment that this should be a either a bare string or at least called `content-type` as the link in the comment was about registry stuff and nothing to do with content-type. The thread in (2) suggests this but I didn't seem to understand the argument that format-spec should be the thing we use

README.md
thomastaylor312: So I missed that when I was digging around. However, I think I need to do some more thinking on the reasoning there. I understand the motivation as described there for not adding it, but based on everything I've seen in real usage of the wasmcloud messaging interface and in my review of most of the common messaging systems, request/reply showed up _everywhere_. Let me do some more reading and come back to this, because, like you mentioned, this should reflect real world usage

thomastaylor312:
Closing in favor of #23 which also incorporates some of the changes from here

PR #23, opened 2024-05-22
feat(*)!: Updates wasi-messaging interface with feedback

thomastaylor312:
This PR incorporates various points of feedback from discussion amongst the interface champions, users of Wasm projects, and discussion in the CNCF wasm WG. This smooths over some of the rough edges and adds support for request/reply paradigms. Some of the bigger changes are:

- Addition of request/reply as an optional interface included in a separate world
- Removal of the custom `format` type in favor of an optional content type string
- Addition of the `topic` field to the `message` type
- Concrete errors in form of a variant


wit/consumer.wit
```diff
@@ -1,17 +1,19 @@
+/// The consumer interface allows a guest to dynamically update its subscriptions and configuration
+/// as well as functionality for completing (acking) or abandoning (nacking) messages.
 interface consumer {
-    // {client, message, channel, error, guest-configuration}
-    use messaging-types.{client, message, channel, error, guest-configuration};
-
-    /// Blocking receive for t-milliseconds with ephemeral subscription – if no message is received, returns None
-    subscribe-try-receive: func(c: client, ch: channel, t-milliseconds: u32) -> result<option<list<message>>, error>;
```
thomastaylor312, line 6: I removed these and replaced them in the new `request-reply` interface as components should handle messages via the `handle` function in `guest` and these are more likely to be used for request/reply operations. If we do want these, let's capture the use cases and probably put them in a separate interface

wit/guest.wit
```diff
@@ -1,10 +1,7 @@
-interface messaging-guest {
-    use messaging-types.{message, guest-configuration, error};
+interface guest {
+    use types.{message, guest-configuration, error};
 
-    /// Returns the list of channels (and extension metadata within guest-configuration) that 
-    /// this component should subscribe to and be handled by the subsequent handler within guest-configuration
-    configure: func() -> result<guest-configuration, error>;
```
thomastaylor312, line 6: Configuration should be provided by the host to the guest via runtime-config or be handled by the messaging implementation itself. This matches how other things are done right now as well in keyvalue

wit/types.wit
```diff
@@ -1,12 +1,21 @@
...
     /// A connection to a message-exchange service (e.g., buffer, broker, etc.).
     resource client {
         connect: static func(name: string) -> result<client, error>;
     }
     
-    /// TODO(danbugs): This should be eventually extracted as an underlying type for other wasi-cloud-core interfaces.
-    resource error {
-        trace: static func() -> string;    
+    /// Errors that can occur when using the messaging interface.
+    variant error {
```
thomastaylor312, line 8: I took a first stab at error cases here. Please feel free to suggest more that might need to be included

wit/request-reply.wit
```diff
@@ -0,0 +1,18 @@
...
+    /// Performs a blocking request/reply operation with an optional timeout. If the timeout value
+    /// is not set, then the request/reply operation will block indefinitely.
+    /// 
+    /// Please note that implementations that provide `wasi:messaging` are responsible for ensuring
+    /// that guests are not allowed to subscribe to channels that they are not configured to
+    /// subscribe to (or have access to). Failure to do so can result in possible breakout or access
+    /// to resources that are not intended to be accessible to the guest. This means implementations
+    /// should validate that the reply-to field is a valid topic the guest should have access to or 
+    /// enforce it via the credentials used to connect to the service.
+    request: func(c: client, msg: message, timeout-ms: option<u32>) -> result<option<list<message>>, error>;
```
danbugs: Do we want an option to block without a timeout?

wit/consumer.wit
```diff
@@ -1,17 +1,19 @@
...
     ///     - checkpointing,
     ///     - etc..
+    /// 
+    /// Please note that implementations that provide `wasi:messaging` are responsible for ensuring
+    /// that guests are not allowed to subscribe to channels that they are not configured to
+    /// subscribe to (or have access to). Failure to do so can result in possible breakout or access
+    /// to resources that are not intended to be accessible to the guest. This means implementations
+    /// should validate that the configured topics are valid topics the guest should have access to or 
+    /// enforce it via the credentials used to connect to the service.
     update-guest-configuration: func(gc: guest-configuration) -> result<_, error>;
```
danbugs: What is the rationale behind leaving this function in the interface considering your comment in `guest.wit`? 

> "Configuration should be provided by the host to the guest via runtime-config or be handled by the messaging implementation itself. This matches how other things are done right now as well in keyvalue"

For clarification - Prior, I had imagined people using this interface to do one of the following:
(1) call `configure` from `guest.wit` to setup a `handler` to receive messages for that specific configuration, or
(2) use the constructs in `consumer.wit` to receive messages from non-commital/ephemeral subscriptions, which was to fulfill the `request-reply` pattern.

W/ adding runtime-config into the picture - how are people expected to set the channel(s) that the handler should be listening to? Plus - isn't runtime-config rn only capable of `get`ting configs? Or, is the idea that the handler will receive all messages and it is the implementor's job to do a `match` on channels (this could be annoying for users because `channel` is just a wrapper for `string`)? If the latter is correct, the design is changing signficantly, because, I had sort of imagined people creating multiple handler implementations for different configs.

wit/request-reply.wit
```diff
@@ -0,0 +1,18 @@
+/// The request-reply interface allows a guest to send a message and await a response. This
+/// interface is considered optional as not all message services support the concept of
+/// request/reply. However, request/reply is a very common pattern in messaging and as such, we have
+/// included it as a core interface.
+interface request-reply {
```
danbugs, line 5: Btw - Do you think this was not fulfilled with the constructs in consumer? Or is the idea mostly here just to split into its' own interface to make it an optional impl while still allow implementing `consumer.wit`?

wit/types.wit
```diff
@@ -16,29 +25,29 @@ interface messaging-types {
...
-        http,
-        amqp,
-        mqtt,
-        kafka,
-        raw
-    }
-  
-    /// A message with a binary payload, a format specification, and decorative metadata.
+    /// A message with a binary payload and additional information
     record message {
```
lukewagner: I wonder if `message` should be a `resource` so that its contents are not always eagerly copied in full (for the same reason as `request` and `response` in WASI HTTP), for the benefit of cases where messages are simply forwarded or only partially inspected.  Also, the `complete-message`/`abandon-message` functions take a `message` by value which again does a full copy -- instead I think these functions want to be methods of a `message` resource (which has the benefit of giving messages a clear identity (for purposes of completion/abandonment) -- as it is, without any sort of "message id" field, I'm not sure how `complete-message`/`abandon-message` would be implemented.

wit/types.wit
```diff
@@ -16,29 +25,29 @@ interface messaging-types {
...
+        /// The topic/subject/channel this message was received or should be sent on
+        topic: channel,
+        /// An optional content-type describing the format of the data in the message. This is 
+        /// sometimes described as the "format" type
+        content-type: option<string>,
+        /// An optional topic for use in request/response scenarios. Senders and consumers of
+        /// messages must not assume that this field is set and should handle it in their code
+        /// accordingly.
+        reply-to: option<string>,
+        /// An opaque blob of data
```
lukewagner: Is the intention to allow the recipient of the message to `reply-to` someone other than the sender of the message?  If "yes", then the blocking `request-reply.request` function seems like a footgun, since it blocks until it gets a reply (which might not come back to it).  If "no", then do we even need a field or can we just say that the `request-reply.request` function automatically sets it on the outgoing side, and then perhaps we could have a `reply` method on the `message` resource (as suggested above) which can be trusted to reply to the correct sender.

wit/guest.wit
```diff
@@ -1,10 +1,7 @@
...
-
-    /// Whenever this guest receives a message in one of the subscribed channels, the message is sent to this handler
+    /// Whenever this guest receives a message in one of the subscribed channels, the message is
+    /// sent to this handler
     handler: func(ms: list<message>) -> result<_, error>;
```
lukewagner: Is wasi-messaging meant to be implemented by messaging systems with reliable (at-most- or at-least-once) delivery?  If `handler` returns an `error`, should the system assume that every `message` in the `list<message>` failed?  But what if some of the `message`s succeeded (finished their work and performed side effects)?  Are failed messages ever retried and, if so, does a component targeting `wasi:messaging` need to be idempotent and, if so, it probably needs some sort of message-id to do that?

wit/types.wit
```diff
@@ -16,29 +25,29 @@ interface messaging-types {
...
     type channel = string;
   
-    /// Configuration includes a required list of channels the guest is subscribing to, and an optional list of extensions key-value pairs 
-    /// (e.g., partitions/offsets to read from in Kafka/EventHubs, QoS etc.).
+    /// Configuration includes a required list of channels the guest is subscribing to, and an
+    /// optional list of extensions key-value pairs (e.g., partitions/offsets to read from in 
+    /// Kafka/EventHubs, QoS etc.).
     record guest-configuration {
         channels: list<channel>,
         extensions: option<list<tuple<string, string>>>
```
lukewagner: It seems like if wasi-messaging has `extensions` that are specially interpreted in important (i.e., not just performance-hinting) ways, then it will produce components that only run correctly on one implementation (say Kafka), which prevents portability.  Could we instead have the motivating use cases (say, partitions/offsets) be explicitly defined by wasi-messaging as separate `interface`s so that a component can declaratively say that it needs those features and different host messaging systems can implement them or not (and if not, maybe they can be virtualized via component).

lukewagner (commented):
I'm only starting to understand the conceptual model of this proposal, so I just had some possibly-naive questions to start with:

wit/types.wit
danbugs: The fact that messaging is used for so many different applications was one of the main pain points I ran into when writing this interface - that is, different possible implementors like: Service Bus, Event Hubs, Kafka, NATS, etc. are all used for different purposes and, with that, have different strengths (e.g., some are used for load balancing, others for decoupling microservices, others for stream processing, and some for telemetry/logging just to name a few applications). For a specific example: Event Hubs does not have the capability to read messages from a consumer group, only send; while, on the other hand, Service Bus has advanced functions to not only read messages, but also peek and so on. With that, my main fear with breaking this interface into smaller pieces like so is that we will just end up with a close-to-zero common denominator and just have interfaces that pretty much represent a service offering (e.g., a “Kafka“ interface). Anyway, a good starting point could be to break messaging into its main applications and use that to bucket different implementors - this way, we can compare their APIs and use that as a blueprint to better type smt like `extensions` and see if that gets us to a point where this interface is sufficiently portable.

danbugs: Also - I had sort of started doing this in here: https://github.com/WebAssembly/wasi-messaging/pull/9#discussion_r1178061562 (i.e., splitting the interface for different applications - e.g., load balancer, but the change got overwritten).

wit/guest.wit
danbugs: > Is wasi-messaging meant to be implemented by messaging systems with reliable (at-most- or at-least-once) delivery?

Yes.

> If `handler` returns an error, should the system assume that every message in the `list<message>` failed?

What do you mean by “failed“? Any error returned by the `handler` is a processing error and should be completely handled by the guest - like, if a handler got a message, that message has been received and it is up to the guest implementor to properly settle it with one of the following methods: https://github.com/WebAssembly/wasi-messaging/blob/c2a7421ad0aece25d11ab5056a8b98f18bedd64a/wit/consumer.wit#L25-L26

That said, I do like the idea of ID-ing messages - I just don‘t know if that is something widely supported at all, implementors might have to create their own implementations for managing IDs if not, which is something we should consider.

wit/types.wit
danbugs: +1 for a `message` `resource`.

wit/request-reply.wit
thomastaylor312: This is documented above. If `timeout-ms` is none, it will block indefinitely

wit/consumer.wit
thomastaylor312: So I left this in after a conversation with @devigned about it. The whole purpose was to handle dynamically indicating to a host that you wanted to update the channels being passed to your handle function especially in cases where you are reacting to another message you've received. I think he can give a better rundown here. 

As for the matching, that is what I intended. I definitely was thinking that a handler function would have a `match` on channels (optionally) if you are subscribing to multiple, or you could just handle each message no matter what the channel. That seems to be a common way to do things in my experience. However, as for setting the channels, that is something that seems solidly in the realm of the host/person running and operating a component. They would say when they start it "please pass all messages from this channel or list of channels to this component." If the component needs to know those channel names, the host can provide them via runtime config. This is also important because otherwise some of the config details could very much be specific to implementations, which should be set outside of the component. Otherwise you end up with the same coupling that exists now

I am curious though how this would work:

> I had sort of imagined people creating multiple handler implementations for different configs

You can only export one `handle` function (or one per aliased interface, e.g. `export handler as another_handler`). So I wasn't sure if you meant that someone would create different components with baked in configs?

wit/request-reply.wit
thomastaylor312: Yeah, this was mainly to make it an optional implementation and simplify the interface for making a request

wit/types.wit
thomastaylor312: I definitely think we should have the complete/abandon functionality be a resource. As for the body of the message, I think having that be behind a resource is good, but I would love to avoid streams if at all possible. Similarly to keyvalue, these values should be small as large values are generally seen as an antipattern. I think both Kafka and NATS have a 1MB default upper limit, but I assume many of the others are also likely to be small. 

With that in mind, would we want a way to "consume" the message when reading the body? I'm not sure what the proper thing here is to do with resources

wit/types.wit
thomastaylor312: I actually like that idea quite a bit. I was wondering if implementations of the request function should just automatically set the reply-to field, but having it be an implementation detail of `request` is probably a good idea.

I probably wouldn't have a `reply` method on the resource because we still want it to be optional, so is there a way to have a `reply` method in `request-reply` that takes a `borrow<message>` in order to send a response?

wit/guest.wit
thomastaylor312: I toyed with the idea of publishing messages and handling messages being a single `message` rather than a list of them. The thought was we could start with that and then add batching, which introduces the complexity mentioned by @lukewagner, as a later thing. Any thoughts?

wit/types.wit
lukewagner: I agree you probably don't need a stream (at least in a 0.2 timeframe).  I'd suggest a resource roughly of the shape:
```wit
resource message {
  topic: func() -> channel;
  ...
  data: func() -> list<u8>;
}
```
(Because we're not trying to stream message bodies, I think we don't need the "may only call once" rules with wasi-http's `body.consume`.)

wit/types.wit
lukewagner: Yes, a `reply` method in `request-reply` could work.  So, iiuc, are you thinking an imported interface like:
```wit
interface request-reply {
  reply: func(reply-to: borrow<message>, reply: message) -> result<_, error>;
}
```
(and I suppose if you call `reply` on a `message` that was not in fact sent via `request` it returns an `error`)?

Thinking about this some more, I have another high-level goals question for you: is it a goal for a runtime to be able to opportunistically link two components together when the sending and receiving component are on the same machine?  This would be like [what we're trying to do in 0.3 with WASI HTTP](https://docs.google.com/presentation/d/1W62iGDiFLSBw6gZU1y8Ec7j0b_5MXOB4Yq448bRr_XQ), but with `wasi:messaging/types.message` in place of `wasi:http/types.{request,response}`.  WASI HTTP has to wait until 0.3 because it's all async/streamy, but with the totally synchronous request-reply design proposed here, we could perhaps get away with it in 0.2.  However, I find that that this goal of "direct component linking" ends up having interesting design implications (which we've incorporated into WASI HTTP) since now you don't have any host code sitting in-between two components (e.g., you can't rely on the host to set the `reply-to` field, since the host isn't sitting between the two components).

thomastaylor312: Yeah that function was exactly what I was thinking. I think direct composition could work as well if the `reply-to` field no longer appears in the message and is handled by the implementation of the resource

wit/guest.wit
lukewagner: Ah, I see: so `complete-message`/`abandon-message` allow you to mark each individual message of a batch; makes sense.  (FWIW, if we `resource`-ify `message` (as discussed above), then we don't need to add any extra "message-id": the handle to the `message` identifies the message and so, e.g., `complete` and `abandon` could be methods on the `message`.)

Regarding batching: if we processed one `message` at a time, then I suppose the return value of the `handle` function could indicate whether the (single) `message` parameter was completed, abandoned, or there was a ephemeral error that wants a retry (which is I suppose is how the host would interpret a trap that kills the instance before it returns?).  That does seem like a simpler API, but maybe it incurs more per-message overhead, but maybe it's fine, at least as a first version?  Some data might help know: does anyone know, for existing message-passing systems, how long it takes to process the average message?  (If it's longer than, say, 1 microsecond, then batching may not be a significant win given the speed of host-to-component calls).

wit/types.wit
lukewagner: Yeah, I can totally imagine there is a wide variety of functionality here.  In the limit, every single messaging implementation could define its own WIT interface (e.g., `apache:kafka`, `azure:service-bus`, etc), and this isn't a bad idea: a customer may wish to use the full feature set of a particular implementation and so an implementation-specific WIT is appropriate (defined and owned by that implementation, not WASI).  Given this option as a baseline, you could say that the role of `wasi:messaging` is to handle cases where a customer intentionally wants portability over multiple implementations (if they didn't, they could just use the WIT for the impl they wanted, and have an altogether better DX).

Given that framing, where the primary value we're adding by defining `wasi:messaging` is abstracting over different implementations, we can (1) provide a least-common-denominator (that works portably over all implementations, but necessarily leaves a bunch of functionality on the table) or (2) start with a least-common-denominator, and add different chunks of optional functionality that are only supported by a (>1-sized) subset of implementations.  (Or start with (1) and then move on to (2).)  I think the real "hard work" of standardization is right-sizing the chunks so that we end up with far fewer chunks than implementations (otherwise, might as well use the impl-specific WITs).

wit/types.wit
lukewagner: The tricky thing I'm having trouble seeing with direct-composition is: if we want direct composition (without magic middleware) we need to have 1 interface that can be both imported and exported (analogous to `wasi:http/handler`; say: `wasi:messaging/handler`).  Once we do this, the natural function signature for the caller (which returns a `message`) must also match that of the callee, which means the callee needs to reply by returning a `message`, not calling an imported `reply` function.  But now this suggests that we need *two* different exportable interfaces: one for request-reply flows (where there is a `message` return value) and one for reply-less messages (with just a `status` return value).  Maybe this is a good thing: I expect handling a request/reply message ends up feeling rather different to a developer than handling a reply-less message and so having different entry points perhaps makes this distinction more clearly in the code (and avoids certain classes of runtime errors).  WDYT?

wit/consumer.wit
danbugs: > As for the matching, that is what I intended. I definitely was thinking that a handler function would have a match on channels (optionally) if you are subscribing to multiple, or you could just handle each message no matter what the channel. That seems to be a common way to do things in my experience.

Ok - If that is how we want to move forward, I would probably just add to the description of the function that we have the expectation that an implementor will `match` or figure out channels themselves.

> However, as for setting the channels, that is something that seems solidly in the realm of the host/person running and operating a component.

About channels, I have always seen it as a consumer responsibility to handle subscriptions - but this may be biased on my use of Service Bus, and MQTT. Also - there is the concept of ephemeral subscriptions to channels too w/ things like `subscribe-try-receive` where a consumer gets a message from a channel and the subscription is scoped only for that one function call. Although, I suppose that doesn't really fit within the `handler` model all that much.

> You can only export one handle function (or one per aliased interface, e.g. export handler as another_handler). So I wasn't sure if you meant that someone would create different components with baked in configs?

This bit was inspired in the idea of maintaining a component's lifetime as short as possible. That is, if we have one handler that manages all channels, the process will probably be very long-lived, which I was trying to avoid. Back in SpiderLightning, the way we added multiple handlers per HTTP endpoint was with a [custom macro](https://github.com/deislabs/spiderlightning/blob/283caa0299750028a0955d1702e8c24d62c415a1/crates/http-handler-macro/src/lib.rs#L26-L54), but that is obviously not portable as it is. So, like you mentioned, this could be done with aliased interfaces - w/ how the interface was previously, you'd have to alias two functions: `configure`, and `handler`. This way, `configure_sasquatch ` would set the subscriptions for `handler_sasquatch`, and so forth.

wit/types.wit
danbugs: That's good framing - I am aligned with you on providing a least-common-denominator interface that can be built upon. And, I suppose we're sort of on the way w/ (2) already by w/ implementing `request-reply` as an optional interface w/ the supporting rationale that it is a common pattern across a wide variety of messaging implementors albeit not supported by all.

lukewagner: Great!  And what's neat about having `request-reply` as an explicitly-declared separate interface is that if you want to run your component on a messaging impl without native `request-reply`, you might be able to virtualize (in the host or in an adapter component) in terms of the underlying primitives that are present since the higher-level expected behavior of `request-reply` is declaratively known.

So then do you think we can remove `extensions`?

danbugs: I'd still leave `extensions` for things like perfomance hinting and alike  - but I do think we can start making an effort to move some of the things that are important and are currently intended for `extensions` (e.g., partition offsets and whatnot) out into their own optional interfaces. @thomastaylor312 , do you want to do this as part of this PR? If not, I can work on it on a follow-up too 👍

lukewagner: I wonder if even performance hints can benefit, portability-wise, by having them explicit named optional fields (of either a record or a resource type (like we did in wasi-http with [`request-options`](https://github.com/WebAssembly/wasi-http/blob/main/wit/types.wit#L331)) so that the meaning of the hint is clear (even if it is ignored).  Or at least this is a conservative way to start, and if we find that there are just too many, then maybe we reconsider.  As a general learning from standards work (many times over on the Web), if you create an open-ended "stick your implementation-specific stuff here" hole, it will end up being used in a way that breaks portability, so it's best to start without it and hold off as long as possible.

wit/guest.wit
thomastaylor312: FWIW we have several people using a custom messaging interface we built for wasmCloud (until this one is ready) in production. It handles and publishes single messages: https://github.com/wasmCloud/messaging/blob/main/wit/messaging.wit. I think it might be worth having batching be a separate interface we define further down the road

wit/types.wit
thomastaylor312: Yeah I can move that stuff out here @danbugs!

wit/types.wit
thomastaylor312: Yeah I think they are handled in different ways and so having the two different interfaces seems to be the way to go if I understood what you said properly

thomastaylor312:
Thanks for all the great feedback! Gimme until tomorrow and I should be able to make all the changes

wit/types.wit
danbugs: Fair play - I'm just not sure how we should balance flexibility and portability in this scenario. I feel like a lot of users might want something like `extensions`, but I also can't disagree that it inevitably hurts portability...

Regardless, I suppose starting with removing it and then adding back functionality per user requests could be the best way to go.

thomastaylor312: So I currently have left extensions in. Given that they are optional and non-structured, I don't think it hurts per se to have some guest hinting something to Kafka, even though they're connected to Event Hubs. Still ok to take it out if we worry about the portability aspect

thomastaylor312:
Ok I've pushed up some additional changes as discussed

danbugs (commented):
These changes are great. Thanks for your work, @thomastaylor312 ! 

Mostly LGTM, just one last note - I'd just probably not delete the `examples.md` document. I think it can be a useful recource for people to see how an interface like this will be used "in the real world" at a quick glance. Do you think we could bring it back or achieve that goal in some other way?

thomastaylor312:
My plan was to actually implement something with this interface and then copy over. I deleted purely because it was completely out of date. I am also fine to put something back together once people are ok with the changes. I'll do that before we merge

wit/types.wit
```diff
@@ -16,29 +25,50 @@ interface messaging-types {
...
-        raw
-    }
-  
-    /// A message with a binary payload, a format specification, and decorative metadata.
-    record message {
-        data: list<u8>,
-        format: format-spec,
-        metadata: option<list<tuple<string, string>>>
+    /// A message with a binary payload and additional information
+    resource message {
```
brooksmtownsend, line 24: I think it would be great to have some additional methods here for adding additional fields to the metadata / overriding the topic in the case of a message proxy

wit/types.wit
```diff
@@ -16,29 +25,50 @@ interface messaging-types {
...
+        constructor(topic: channel, data: list<u8>, content-type: option<string>, metadata: option<list<tuple<string, string>>>);
+        /// The topic/subject/channel this message was received or should be sent on
+        topic: func() -> channel;
```
brooksmtownsend: Looking through the types, naming of parameters, and comments, I'm getting a little turned around with the difference between a topic/subject/channel. Taking a look at pubsub messaging systems, I know that Apache Kafka, Google Cloud Pub/Sub, and Confluent use "topic", NATS uses "subject" but also uses that interchangeably with "topic", and then Redis uses channels. These are just a few options and I can totally see where the naming here came from.

This might be a bikeshed, but what are your thoughts on just calling this a "topic" all around the interface, and then clarifying on the type itself, e.g.
```
    /// Topics, aka subjects, channels, groups in pubsub systems
    /// ...
    type topic = string,
```

Since it's a strawman I'm fine considering any terminology here, just suggesting to keep it consistent

brooksmtownsend (commented):
Overall looks nice. The new request-reply interface looks really great! Left some comments for discussion

wit/types.wit
lukewagner: Are there any specific high-value examples of hints you can point to which would make the first release of this interface non-viable?  If so, again, maybe they could be optional fields of a record/resource; if not, then perhaps, just based on the general 80/20-rule principle that you've been using, we should leave if off for now (we can always add, but much harder to subtract).

wit/types.wit
thomastaylor312: I think topic would be a lot clearer personally. Any thoughts @danbugs since I think you might have been the one who picked "channel"

wit/types.wit
thomastaylor312: I personally don't have any, but I also don't use Kafka a bunch, which is what I think has been stated here. @danbugs and @devigned do either of you have any strong examples here? Otherwise I'm inclined to agree with @lukewagner and take it out

wit/types.wit
danbugs: Iirc, I just chose `channel` to avoid picking a specific nomenclature 👍

imports-request-reply.md
```diff
@@ -0,0 +1,268 @@
...
+<p>The request-reply interface allows a guest to send a message and await a response. This
+interface is considered optional as not all message services support the concept of
+request/reply. However, request/reply is a very common pattern in messaging and as such, we have
+included it as a core interface.</p>
+<hr />
+<h3>Types</h3>
+<h4><a name="client"></a><code>type client</code></h4>
+<p><a href="#client"><a href="#client"><code>client</code></a></a></p>
+<p>
+#### <a name="message"></a>`type message`
```
devigned, line 148: These `#### ...` are not rendering well in the markdown.

examples.md
```diff
@@ -1,178 +0,0 @@
-# Examples
```
devigned, line 1: It would be helpful to see some examples of usage with the changes to the interface.

wit/consumer.wit
```diff
@@ -1,27 +1,17 @@
+/// The consumer interface allows a guest to dynamically update its subscriptions and configuration
 interface consumer {
```
devigned, line 2: Are the guest and consumer interfaces separate due to only wanting to export the handler function and not both handler and update-guest-configuration?

wit/types.wit
```diff
@@ -16,29 +25,50 @@ interface messaging-types {
...
-    }
-  
-    /// A message with a binary payload, a format specification, and decorative metadata.
-    record message {
-        data: list<u8>,
-        format: format-spec,
-        metadata: option<list<tuple<string, string>>>
+    /// A message with a binary payload and additional information
+    resource message {
+        constructor(topic: channel, data: list<u8>, content-type: option<string>, metadata: option<list<tuple<string, string>>>);
```
devigned: If `send` takes as a parameter `channel` what is the significance of having `channel` on the constructor for a message? If I create a `message` with a channel and call `send` with a different `channel`, what should happen?

wit/types.wit
devigned, line 24: Many of the messaging implementations have the idea of a Time to Live (ttl) on a message. For example, upon receiving a message for processing, the ttl would be specified on the message metadata and the broker will lock the message in the queue so no other consumer can process the message. To ensure the message is not held forever, the broker will time bound the activity with a ttl. If the message is not ack'd in that time period, the broker will release the lock on the message and allow another consumer to receive the message.

How would a consumer discover the ttl? Would it be in the metadata? Is it up to each implementor to define how metadata is represented / interpreted?

devigned (commented):
I like where this is going. I have a handful of comments and questions.

wit/consumer.wit
thomastaylor312: Now that you say it, they should probably just be combined

wit/types.wit
thomastaylor312: I did add documentation for this on the `send` method about what happens. The use case there I think was to override it if you're forwarding on a message. However, I think we should just add a method for modifying the channel and then you change it on the message itself

imports-request-reply.md
thomastaylor312: These are auto generated so there isn't much I can do there

examples.md
thomastaylor312: Yeah, once we're in agreement with things, I am going to update this

wit/types.wit
thomastaylor312: Normally I've seen the ttl only when you're doing the more complex patterns with things like work queues and anything else that uses delivery guarantees. As far as I understood, we're not covering those patterns right now. I think anyone can put whatever they want on metadata, but having the abandon/complete message is enough to provide an acknowledgement. Right now this has to be generic enough to support basic eventing which requires no acknowledgement or guarantees, and anything can subscribe to it. I think one of the first things I'd want to add once we get implementations for this in place would be to create a new interface (or messaging-extension package) that has extension methods for the `message` allowing you to fetch things like TTL and other details for guaranteed delivery

For request-reply, we do support a timeout for an operation so that is supported as well

wit/consumer.wit
thomastaylor312: Oh wait, that's right. They are separate because a host exports the configuration function to the guest, whereas the guest exports the handler to the host

wit/types.wit
thomastaylor312: I've kept it channel for now. Let me know if there are any strong opinions to the contrary

wit/producer.wit
```diff
@@ -1,5 +1,8 @@
+/// The producer interface is used to send messages to a channel/topic.
 interface producer {
-    use messaging-types.{client, channel, message, error};
+    use types.{client, channel, message, error};
     
-    send: func(c: client, ch: channel, m: list<message>) -> result<_, error>;
+    /// Sends a message to the given channel/topic. If the channel/topic is not empty, it will
+    /// override the channel/topic in the message.
+    send: func(c: client, ch: channel, m: message) -> result<_, error>;
```
yordis:
```suggestion
    send: func(c: client, m: message) -> result<_, error>;
```

I noticed the `message` has a `topic` value in the constructor.

Would make sense to remove `topic` parameter and rely on the `message`?

I am not sure what the value of having two places to read the `topic` value from is.

yordis:
```suggestion
    send: func(c: client, ch: channel, m: message, opts: option<send-options>) -> result<_, error>;
```

```wit
record send-options {
  timeout-ms: option<u32>
}
```

I expect the timeout to be configurable. How are you envisioning the situation around it?

Are you envisioning to add the value to the `message` itself?

wit/request-reply.wit
```diff
@@ -0,0 +1,16 @@
...
+    use types.{client, message, error};
+
+    /// Performs a blocking request/reply operation with an optional timeout. If the timeout value
+    /// is not set, then the request/reply operation will block indefinitely.
+    request: func(c: client, msg: message, timeout-ms: option<u32>) -> result<option<list<message>>, error>;
```
yordis:
```suggestion
    request: func(c: client, msg: message, options: option<request-options>) -> result<option<list<message>>, error>;
```

```wit
record request-options {
  timeout-ms: option<u32>
}
```

I suggest avoiding `option` as the function's signature to avoid breaking changes from the caller and instead use an object as a third parameter

wit/request-reply.wit
```diff
@@ -0,0 +1,16 @@
...
+
+    /// Replies to the given message with the given response message. The details of which channel 
+    /// the message is sent to is up to the implementation. This allows for reply to details to be
+    /// handled in the best way possible for the underlying messaging system.
+    reply: func(reply-to: borrow<message>, reply: message) -> result<_, error>;
```
yordis, line 42:
```suggestion
    reply: func(reply-to: borrow<message>, reply: message, opts: option<reply-options>) -> result<_, error>;
```

```wit
record reply-options {
  timeout-ms: option<u32>
}
```

Same over here. How do we deal with the timeout config?

wit/types.wit
jocrau: IMHO the term `channel` implies a "transport" from one end to the other. The alternatives `topic` and `subject` are neutral to that aspect. But they have the advantage that they imply to "address" the message source. The messaging (at least in the IIoT space) is organized around an "asset" not the "channel" the messages are sent through. So, I prefer `topic` or `subject`.

wit/types.wit
```diff
@@ -16,29 +25,56 @@ interface messaging-types {
...
+        /// 
+        /// A message can exist under several statuses:
+        /// (1) available: the message is ready to be read,
+        /// (2) acquired: the message has been sent to a consumer (but still exists in the queue),
+        /// (3) accepted (result of complete): the message has been received and ACK-ed by a consumer and can be safely removed from the queue,
+        /// (4) rejected (result of abandon): the message has been received and NACK-ed by a consumer, at which point it can be:
+        ///         - deleted,
+        ///         - sent to a dead-letter queue, or
+        ///         - kept in the queue for further processing.
+        abandon: func() -> result<_, error>;
```
yordis:
```suggestion
        abandon: func(abandon-msg) -> result<_, error>;
```

```wit
record abandon-msg {
  reason: ...
}
```

I don't know if it would be worth adding some reason for abandonment. I'm just generalizing my thoughts here to see what sticks.
These would be friendly messages in the `Competing Consumer` model. 

Maybe it is not worth having it as a core functionality; people could add it as an add-on.

Action | Description
-- | --
Unknown | The client does not know what action to take. Let the server decide.
Park | Park the message and do not resend. Put it on poison queue.
Retry | Explicitly retry the message.
Skip | Skip this message do not resend and do not put in poison queue.

wit/types.wit
yordis: Wouldn't it be better to remove the channel from `send` instead?

wit/types.wit
danbugs: I'm good w/ taking it out 👍

wit/consumer.wit
devigned: Thank you for the clarification. Just wanted to make sure I understood the intent.

wit/types.wit
devigned: I would imagine that `channel` should have `send(message)`.

wit/types.wit
devigned: sgtm

wit/types.wit
devigned: AMQP structures related to this are here: http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-rejected.

wit/types.wit
thomastaylor312: Yep, I meant to do this and forgot to in my last commit. Will do shortly

wit/producer.wit
thomastaylor312: See [this thread above](https://github.com/WebAssembly/wasi-messaging/pull/23#discussion_r1627878735) for more information around ttl/timeout. Right now this only matters if you're in a request/reply

wit/request-reply.wit
thomastaylor312: A timeout would only matter here if you're expecting the thing you sent the reply to respond with a confirmation. More complex patterns like what Kafka Streams and NATS JetStream provide are going to be layered on top of this. Those types of patterns can handle things like acks and double_acks. Everything in this package, with the exception of request reply (where the request should have an eventual timeout), is doing basic pub sub with no explicit guarantees around deliverability

wit/request-reply.wit
thomastaylor312: With the inclusion of the new `@since` stuff, this is probably the way to go. I'll make the change

wit/types.wit
thomastaylor312: What do we think about having this just be a string rather than a specific state? I do like having additional context for sure, but am trying to keep this as simple as possible

wit/types.wit
thomastaylor312: Those sound like great reasons to me. I've updated this to be `topic`

wit/request-reply.wit
```diff
@@ -0,0 +1,27 @@
...
+/// included it as a core interface.
+interface request-reply {
+    use types.{client, message, error};
+
+    /// Options for a request/reply operation
+    record request-options {
+        /// The maximum amount of time to wait for a response. If the timeout value is not set, then
+        /// the request/reply operation will block until a message is received in response.
+        timeout-ms: option<u32>,
+        /// The number of expected replies. If the number of replies is not set, then the
```
thomastaylor312: @danbugs @devigned As I was adding the new request options, I thought about how we want to handle this. Currently, `request` returns a list of messages. If we are returning a list of messages, then it makes sense we might want to specify (optionally) the number of replies we expect for things like scatter/gather operations. I am not totally sold on the idea, but put it in here so it was easy to grok. I think there are some options here:

- Keep support for receiving multiple responses and have this optional `expected-replies` field (more complex, but more flexible in what you can use it for)
- Make the `request` method only return a single message (this is the simplest, but does limit you to more of a client-server type pattern)
- Maybe something in between? I can't quite think of one, but I don't have a monopoly on thoughts around messaging 😅

wit/types.wit
yordis: @thomastaylor312 yeah, that is why I left the type out.

wit/request-reply.wit
yordis: > A timeout would only matter here if you're expecting the thing you sent the reply to respond with a confirmation.

When you say "reply to respond with a confirmation," what do you have in mind?

I am focused more on network timeout; the guarantee is about the system, at the very least, getting the message.

wit/request-reply.wit
```diff
@@ -0,0 +1,27 @@
...
+        /// request/reply operation will block until a single message is received in response (or a
+        /// timeout value is reached). This can be used in conjunction for "scatter/gather" type
+        /// operations.
+        expected-replies: option<u32>,
```
yordis: That feels very specific to a given implementation.

Could the implementation of scatter/gather be possible without it? From the spec perspective, it should be enough as long as I can get multiple replies to a given message.

thomastaylor312: Like I mentioned [here](https://github.com/WebAssembly/wasi-messaging/pull/23#discussion_r1639029618), I'm not totally sold on it, but pushed it up so it would be easier for discussion. I'll resolve this thread here and then we can continue in the other discussion

wit/types.wit
```diff
@@ -1,44 +1,77 @@
...
-    }
-  
-    /// A message with a binary payload, a format specification, and decorative metadata.
-    record message {
-        data: list<u8>,
-        format: format-spec,
-        metadata: option<list<tuple<string, string>>>
+    /// A message with a binary payload and additional information
+    resource message {
+        constructor(topic: topic, data: list<u8>, content-type: option<string>, metadata: option<list<tuple<string, string>>>);
```
yordis: What are the expectations around the Message ID? Such metadata is so critical that it would be good to have one way to figure this out.

yordis:
```suggestion
        constructor(topic: topic, data: list<u8>, opts: option<message-options>);
```

```wit
record message-options {
  content-type: option<string>
  metadata: option<list<tuple<string, string>>>
}
```

I feel it will be prudent to avoid breaking changes in the function signature

thomastaylor312: Is a message ID actually important to a guest here? Most of the time those are used in acks or other advanced operations. Each implementation can take the wit type `message` and convert it to/assign it an ID based on implementation details IMO.

thomastaylor312: @lukewagner I'm curious on your thoughts here on how constructors should be handled. In particular:

1. Is it idiomatic to have another record type you use in a constructor?
2. Are there any other considerations around breaking a constructor API that we should take into account here? At this point I don't think the `@since` and `@unstable` would necessarily help here

yordis: From the guest's perspective, I think it is critical for the `guest` to control the identity of the message, the `guest` is most likely the component aware of the domain, while the `host` is at the platform level.

What the `host` would do with it shouldn't be part of the spec. That is a different story. Still, `guest` could find a component that provides the guarantees required around the dedupe and whatnot.

It's worth saying that the ID doesn't mean unique; it means an identifier.

wit/producer.wit
yordis: I wasn't thinking of the `ttl` message itself, but the network timeout, if you will, something around the lines of "I give you 2 seconds to tell me you have received (not process) the message" per request-based

wit/request-reply.wit
danbugs: > Make the request method only return a single message (this is the simplest, but does limit you to more of a client-server type pattern)

I think this would just cause people to make wrappers around the single reply mechanism for multiple-reply, so I'd avoid it.

Now, going over your other suggestion, I think the main question we should be asking is: What if the implementor doesn't really offer support for said feature? How should the implementor behave if, regardless, the field is set? And I'm not just saying this about `expected-replies`, but other `opts` too like `timeout-ms`. Rather than having options all together, I'm debating if we wouldn't be better off w/ entirely different functions, so that an implementor can make an explicit choice about implementing them or not. What are your thoughts, @thomastaylor312 ?

thomastaylor312: If people wanted to opt out of implementations, each one of those functions would have to be a separate interface, which could get messy. I also feel like that no matter where each of those functions live, it would start to get really annoying to have to find the right function for each thing you're doing, but that is more of a minor gripe to me than the former. So my initial thought is that I don't quite think that would work.

The comment about options is interesting. A timeout value is simple enough that even if a client doesn't natively support it, it is really easy to add that via code. So I feel like that is fairly easy. I could say the same about expected replies, but I can also start to see the slippery slope here. At what point do additional options cross over into "yeah that is going to require custom logic for every implementation." Anyway, if my first answer is correct in that many functions would be a bad idea (which it could not be), then the best path forward would be to use options but to add even more documentation that we must be very careful adding anything else to it

wit/types.wit
yordis: For whatever is worth, "messaging.message_id" OpenTelemetry Semantic Conventions Trace exists. Just to point to the important of such information

wit/types.wit
```diff
@@ -1,44 +1,77 @@
...
+    type topic = string;
   
-    /// Configuration includes a required list of channels the guest is subscribing to, and an optional list of extensions key-value pairs 
-    /// (e.g., partitions/offsets to read from in Kafka/EventHubs, QoS etc.).
-    record guest-configuration {
-        channels: list<channel>,
-        extensions: option<list<tuple<string, string>>>
+    /// Configuration includes a required list of channels the guest is subscribing to
+    record config {
+        channels: list<topic>,
```
jocrau: Now that the element is renamed to `topic`, shouldn't the list be named `topics` for consistency?

thomastaylor312: Oh yep. Good catch. Thank you!

wit/consumer.wit
```diff
@@ -1,27 +1,17 @@
...
-    complete-message: func(m: message) -> result<_, error>;
-    abandon-message: func(m: message) -> result<_, error>;
+    /// 
+    /// Please note that implementations that provide `wasi:messaging` are responsible for ensuring
+    /// that guests are not allowed to subscribe to channels that they are not configured to
+    /// subscribe to (or have access to). Failure to do so can result in possible breakout or access
+    /// to resources that are not intended to be accessible to the guest. This means implementations
+    /// should validate that the configured topics are valid topics the guest should have access to or 
+    /// enforce it via the credentials used to connect to the service.
+    update-config: func(gc: config) -> result<_, error>;
```
lukewagner: I think readability might be enhanced by renaming this function to something like `set-subscriptions: func(topics: list<string>) -> result<_, error>;` (and deleting `config`).  Then, if we want to add other kinds of configuration in the future, we can just add other functions (which is a semver compatible change, unlike, atm, adding fields to records).  Also, do we have to worry about the list of topics growing large, and then each time you add 1 topic, it's an O(n) operation?  If so, there could be `add-subscription`, `remove-subscription`, etc.  But it's also fine to wait and see if this is a problem.

wit/guest.wit
```diff
@@ -1,10 +1,8 @@
...
-    /// Returns the list of channels (and extension metadata within guest-configuration) that 
-    /// this component should subscribe to and be handled by the subsequent handler within guest-configuration
-    configure: func() -> result<guest-configuration, error>;
-
-    /// Whenever this guest receives a message in one of the subscribed channels, the message is sent to this handler
-    handler: func(ms: list<message>) -> result<_, error>;
+    /// Whenever this guest receives a message in one of the subscribed channels, the message is
+    /// sent to this handler. The guest is responsible for matching on the channel and handling the
+    /// message accordingly.
+    handler: func(ms: message) -> result<_, error>;
```
lukewagner: Nit: to be symmetric with `wasi:http`, how about renaming the `guest` interface to `incoming-handler`, and renaming the `handler` function to `handle`?

wit/types.wit
```diff
@@ -1,44 +1,77 @@
...
+        /// 
+        /// A message can exist under several statuses:
+        /// (1) available: the message is ready to be read,
+        /// (2) acquired: the message has been sent to a consumer (but still exists in the queue),
+        /// (3) accepted (result of complete): the message has been received and ACK-ed by a consumer and can be safely removed from the queue,
+        /// (4) rejected (result of abandon): the message has been received and NACK-ed by a consumer, at which point it can be:
+        ///         - deleted,
+        ///         - sent to a dead-letter queue, or
+        ///         - kept in the queue for further processing.
+        complete: func() -> result<_, error>;
```
lukewagner: Given that `handler` now takes a single message, could we remove the `complete`/`abandon` methods from the `message` resource and instead use the return value of `handler` to indicate completion/abandonment?  This has a few advantages:
* it avoids the cross-product of spec-questions of what happens if the message was { completed, abandoned, neither } x and the result of `handler` was { `ok`, `error`, trapped }, all of which otherwise need precise answers
* it avoids questions of how `complete` and `abandon` interact with `request-reply`
* it's more like normal success/failure control flow

wit/request-reply.wit
```diff
@@ -0,0 +1,27 @@
+/// The request-reply interface allows a guest to send a message and await a response. This
+/// interface is considered optional as not all message services support the concept of
+/// request/reply. However, request/reply is a very common pattern in messaging and as such, we have
+/// included it as a core interface.
+interface request-reply {
+    use types.{client, message, error};
+
+    /// Options for a request/reply operation
+    record request-options {
```
lukewagner: I'd suggest copying [`request-options` in wasi-http](https://github.com/WebAssembly/wasi-http/blob/main/wit/types.wit#L331) and making this a resource, with methods to set/get each field.  This has the advantage of allowing semver-compatible field additions over time.

wit/request-reply.wit
```diff
@@ -0,0 +1,27 @@
...
+        timeout-ms: option<u32>,
+        /// The number of expected replies. If the number of replies is not set, then the
+        /// request/reply operation will block until a single message is received in response (or a
+        /// timeout value is reached). This can be used in conjunction for "scatter/gather" type
+        /// operations.
+        expected-replies: option<u32>,
+    }
+
+    /// Performs a blocking request/reply operation with an optional set of request options
+    request: func(c: client, msg: message, opts: option<request-options>) -> result<option<list<message>>, error>;
```
lukewagner: I think you need `c` to be a `borrow<client>`.  Also, it seems like `expected-replies` is a pretty essential part of how this function works (it explains the `list` in the return type) so maybe it should be a proper parameter instead of being "hidden" in `request-options`?  It sounds like maybe it's not common to use a value other than 1, so what about improving the ergonomics of the common case with:
```
request: func(c: borrow<client>, m: message, opts: option<request-options>) -> result<mesage, error>;
request-list: func(c: borrow<client>, m: message, expected-replies: u32, opts: option<request-options>) -> result<list<message>, error>;
```

wit/request-reply.wit
```diff
@@ -0,0 +1,27 @@
...
+
+    /// Replies to the given message with the given response message. The details of which channel 
+    /// the message is sent to is up to the implementation. This allows for reply to details to be
+    /// handled in the best way possible for the underlying messaging system.
+    reply: func(reply-to: borrow<message>, reply: message) -> result<_, error>;
```
lukewagner, line 42: What if instead of `reply` being an imported interface, it was instead an exported interface:
```wit
interface incoming-request-reply-handler {
  handle: func(msg: message) -> result<message, error>;
}
```
Here, the reply is just the success case of the `result`.  This can avoid some of the weird spec interaction questions like I was pointing out in my other comment on `complete`.  Thus, there could be a request-reply world that looks like:
```wit
world request-reply-messaging {
  include messaging;
  import outgoing-request-reply-handler;
  export incoming-request-reply-handler;
}
```

wit/types.wit
```diff
@@ -1,44 +1,77 @@
...
   
-    /// There are two types of channels:
-    /// - publish-subscribe channel, which is a broadcast channel, and
-    /// - point-to-point channel, which is a unicast channel.
-    ///
-    /// The interface doesn't highlight this difference in the type itself as that's uniquely a consumer issue.
-    type channel = string;
+    /// The topic of a message. This is also called subject, channel, or group in various messaging
+    /// systems. What this value means is up to the implementation.
+    type topic = string;
```
lukewagner: I've generally found that type definitions like this that assign a name to `string` mostly just add indirection that the reader has to chase down without providing any real abstraction (e.g., it's a semver breaking change to attempt to change `topic` to equal any other type, and generated bindings will fully expose the fact that `topic` is actually a string).  So I'd suggest deleting `topic` and relying on the parameter/field names to convey topic-ness.

wit/types.wit
lukewagner: To carve out a way to add optional fields in a semver-compatible way (at least in a 0.2 timeframe before we relax subtyping), you can use a resource type with methods for each field, which is what [wasi-http does with `request-options`](https://github.com/WebAssembly/wasi-http/blob/main/wit/types.wit#L331).

lukewagner (commented):
Changes are looking good!  Just a few more ideas/suggestions:

wit/types.wit
thomastaylor312: I'm glad you brought this up. I actually started doing it this way and then was unsure if it was a good idea. Now that it also came from someone else I feel more confident 😅

wit/request-reply.wit
thomastaylor312: So I actually think this would make things too complicated. This would make it so a component would have implement a different `handle` function if it is responding to requests. The `reply` function here is for a handler to use to reply to a message it has received, no matter the source. The host could throw it away or decide not to send it (for example, in the code I've written for NATS, if reply wasn't set, we just don't send anything back). That could still work here, but I don't like the idea of having another interface the guest has to export rather than just sending the reply and letting the implementation decide what to do with it (which feels more compatible)

wit/types.wit
thomastaylor312: I think I'd prefer to hold off and add this once we have a concrete use case for it

wit/producer.wit
thomastaylor312: I don't think we'll want to add that kind of guarantee to the interface right now. That is much more of a Kafka Stream or NATS JetStream type functionality. The guarantees we offer here are just that the host responds with an error if unable to send a message out, not if it was received. Like I mentioned elsewhere, I think that will be an extension interface on top of this

thomastaylor312:
Hey all! Sorry for the delay in responding, got pulled away to some other work. I've implemented all suggested changes and I think we're ready for a final review @danbugs @devigned @lukewagner. Once we've approved this, I'll add back an examples file that matches the new interfaces before merging

wit/types.wit
```diff
@@ -1,44 +1,46 @@
...
-    }
-  
-    /// A message with a binary payload, a format specification, and decorative metadata.
-    record message {
-        data: list<u8>,
-        format: format-spec,
-        metadata: option<list<tuple<string, string>>>
+    /// A message with a binary payload and additional information
+    resource message {
+        constructor(topic: string, data: list<u8>);
```
thomastaylor312, line 25: Please note that I shrunk this down to avoid any breaking changes. I can't think of any scenario where you wouldn't need to provide at least a topic and a body (though sometimes that body would be empty, so I could get behind an argument we don't need that parameter, but I still think it is useful to have). You can still set the other data members using the getters and the setters

wit/request-reply.wit
lukewagner: Maybe you could help me understand how these request-reply messaging APIs are used in practice because it feels like there's something I'm missing.  I guess a basic question is: is calling the `reply` method on a `message` passed as the parameter to `handle` able to express some use cases that returning the reply as a return value cannot?  If so: what are they?  If not, how is ye olde parameters+results not the simpler interface?

If there is a good reason for having `reply` be a function, I think the spec text needs to go into more detail about what happens in various cases such as:
* What is the interaction between the return value of `handle` (the `result`) and `reply`: if I `reply` but then return an `error` (signalling that this message should not be considered "processed", per the current spec text), does that mean `handle` might get called again for the same `message` and, in that case, will the host care of making sure there is only one `reply` or does that guest have to worry about that?
* If I destroy a `message` without calling `reply`, is some default reply sent (that would be received by a blocking `request` caller) and if so, what is it?
* What happens if I call `reply` multiple times for the same `message`?
* If I `reply` to a `message` that was sent by `send` (not `request`) what happens?  Is there any way to detect whether a `message` is "replyable"?
* If I forward a `message` (to any of `send`, `request` or `reply`) and *that* recipient calls `reply`, who does the reply go to?

I think all these questions go away if replies are return values, because there are just less moving parts (which suggests that it is indeed the simpler interface).

thomastaylor312:
Please note that I just pushed one more commit (so we can pop it off if people hate it) to change how `request-multi` works. One of the uses of `request-multi` is to support a scatter/gather operation. In these cases, you might not know how many requests you are going to receive, so you can't set expected replies. Generally these wait until timeout and then return the results. This commit adds the ability to support all the different use cases for request-multi by adding expected replies to the request options. We don't have to support this, but when I was reviewing different use cases one more time, I was reminded of the scatter/gather operation. Just let me know if people hate it or if it isn't documented clearly

wit/consumer.wit
```diff
@@ -1,27 +1,14 @@
...
+
+    /// Set the current subscriptions for this guest.
+    /// 
+    /// Please note that implementations that provide `wasi:messaging` are responsible for ensuring
+    /// that guests are not allowed to subscribe to channels that they are not configured to
+    /// subscribe to (or have access to). Failure to do so can result in possible breakout or access
+    /// to resources that are not intended to be accessible to the guest. This means implementations
+    /// should validate that the configured topics are valid topics the guest should have access to or 
+    /// enforce it via the credentials used to connect to the service.
+    set-subscriptions: func(topics: list<string>) -> result<_, error>;
```
yordis, line 13:
```suggestion
    subscribe: func(topics: list<string>) -> result<_, error>;
```

Maybe? Calling it a "setter" feels weird

wit/request-reply.wit
```diff
@@ -0,0 +1,43 @@
...
+    request: func(c: borrow<client>, msg: message, opts: option<request-options>) -> result<message, error>;
+
+    /// Performs a blocking request/reply operation with an optional set of request options. This
+    /// returns all replies received up until timeout or the configured set of expected replies. It
+    /// is recommended to use a `request-options` with the timeout set to ensure that the operation
+    /// does not block indefinitely. Unlike request, this function should not return an error on
+    /// timeout and should instead return all of the replies received up to that point. This is to
+    /// faciliate use in scatter/gather operations where the number of expected replies is not
+    /// known.
+    request-multi: func(c: borrow<client>, msg: message, opts: option<request-options>) -> result<list<message>, error>;
```
yordis, line 37:
```suggestion
    request-multi: func(c: borrow<client>, msg: message, opts: option<request-multi-options>) -> result<list<message>, error>;
```

I would love different data structures here

yordis (changes requested):

thomastaylor312:
@lukewagner and I had a good convo about the reply stuff today as well as what the portability criteria should be. I am going to noodle on that a bit and also probably get some feedback from the Wasm WG in CNCF (since they've been interested in this) on the possible options. So I might wait to review at least the request/reply stuff until we can have some conversations

wit/guest.wit
```diff
@@ -1,10 +1,10 @@
...
 
-    /// Returns the list of channels (and extension metadata within guest-configuration) that 
-    /// this component should subscribe to and be handled by the subsequent handler within guest-configuration
-    configure: func() -> result<guest-configuration, error>;
-
-    /// Whenever this guest receives a message in one of the subscribed channels, the message is sent to this handler
-    handler: func(ms: list<message>) -> result<_, error>;
+    /// Whenever this guest receives a message in one of the subscribed channels, the message is
+    /// sent to this handler. The guest is responsible for matching on the channel and handling the
+    /// message accordingly. An Ok result indicates that the message was handled successfully and
```
danbugs, line 6:
```suggestion
    /// message accordingly. An `Ok` result indicates that the message was handled successfully and
```

~nit

wit/request-reply.wit
```diff
@@ -0,0 +1,43 @@
+/// The request-reply interface allows a guest to send a message and await a response. This
+/// interface is considered optional as not all message services support the concept of
```
danbugs, line 2:
```suggestion
/// interface is considered optional as not all messaging services support the concept of
```
~nit

wit/request-reply.wit
danbugs: **In reply to the parallels between `handle`'s `result` and `reply` + your two first bullet points:** The way I see it, `wasi:messaging`'s `handle` is similar to `wasi:http`s `incoming-handler`'s `handle`. So, maybe, just like it, we shouldn't have a `return` type here. Instead, the `guest` should either do nothing or explicitly `reply` to a given message. Otherwise, unavoidably, I think we will need to have an implicit reply and we should move way from anything implicit like that.

> What happens if I call `reply` multiple times for the same message?

First, I think we have to better define who the reply is to. That is, I don't think this should be up to the implementation like it currently says - Considering `request-reply` is an imported interface (and, hence, implemented by a host), why does the, say, `NATS` implementor decide who I am sending messages to? Instead, I think we should be clear that it goes to the `topic` specified in the `reply` `message`. Note, it should not go to the `topic` in the `reply-to` message because, if it did, the message would be `handle`d by the same `handle`r we are calling `reply` from and we'd get infinite recursion.

Second, when `reply`ing, the consumer is essentially acting as a producer, so, whatever happens when we reply multiple times should be decided by the `handle`r of whatever topic that message was sent to.

> If I `reply` to a message that was sent by `send` (not `request`) what happens? Is there any way to detect whether a message is "replyable"?

There should be no difference. Every message should be replyable. Maybe we should clarify this, if you think this could be a point of confusion.

> If I forward a message (to any of `send`, `request` or `reply`) and that recipient calls reply, who does the reply go to?

I touched on this point in my previous reply. Currently, the interface states that this should be decided by the implementor of reply (i.e., a host), but I don't think that should be the case. Instead, we should always reply the topic specified in the `reply` `message`, so it can be `handle`d by some other `handle`r.

wit/request-reply.wit
```diff
@@ -0,0 +1,43 @@
...
+    /// returns all replies received up until timeout or the configured set of expected replies. It
+    /// is recommended to use a `request-options` with the timeout set to ensure that the operation
+    /// does not block indefinitely. Unlike request, this function should not return an error on
+    /// timeout and should instead return all of the replies received up to that point. This is to
+    /// faciliate use in scatter/gather operations where the number of expected replies is not
+    /// known.
+    request-multi: func(c: borrow<client>, msg: message, opts: option<request-options>) -> result<list<message>, error>;
+
+    /// Replies to the given message with the given response message. The details of which channel 
+    /// the message is sent to is up to the implementation. This allows for reply to details to be
```
danbugs, line 40: I touched on this in my reply to Luke in another comment, but I don't think the details of which channel the message is sent to should be up to the implementation. Considering `request-reply` is an `imported` interface, it will be implemented by a host - say, `NATS`. Why should `NATS` decide where my messages go? Instead, I think we should be clear the message should go to the topic in the `reply` `message`.

wit/guest.wit
```diff
@@ -1,10 +1,10 @@
...
+    /// should be considered processed. If explicitly abandoning the message, the guest should return
+    /// error::abandoned with a message indicating why.
+    handle: func(ms: message) -> result<_, error>;
```
danbugs, line 9: I talked about this in my reply to Luke in another comment, but I believe the `handle` function should maybe no longer have a return type. Instead, implementors (i.e., guests) should either: (1) do nothing with a received message, or (2) explicitly `reply`. 

Otherwise, I think we can't get away from implicit replies - right? 'Cause who else (other than a host) could be looking at the return type of `handle`?

However, note that, with `request-reply` being an optional interface, the `handle`r may have to instead use `send` over `reply` if `request-reply` is unimplemented.

wit/messaging.wit
```diff
@@ -5,7 +5,12 @@ world imports {
 	import consumer;
 }
 
+world imports-request-reply {
+	include imports;
+	import request-reply;
+}
+
```
danbugs, line 12: Do we maybe also want a world where we `import-request-reply` and `export incoming-handler`?

```suggestion
world imports-request-reply {
	include imports;
	import request-reply;
}

world messaging-with-request-reply {
	include imports-request-reply;
	export incoming-handler;
}
```

wit/types.wit
```diff
@@ -1,44 +1,46 @@
...
     }
     
-    /// TODO(danbugs): This should be eventually extracted as an underlying type for other wasi-cloud-core interfaces.
-    resource error {
-        trace: static func() -> string;    
+    /// Errors that can occur when using the messaging interface.
+    variant error {
+        /// The requested option is not authorized. This could be a topic it doesn't have
+        /// permission to subscribe to, or a permission it doesn't have to perform a specific
+        /// action. This error is mainly used when calling `set-subscriptions` on a guest.
```
danbugs, line 11:
```suggestion
        /// The requested option is not authorized. 
        /// This could be because a topic was not permissive to being
        /// subscribed to, or because permissions were lacking to
        /// perform some other specific action. This error is mainly used 
        /// when calling `set-subscriptions` on a guest
```
~nit

wit/types.wit
```diff
@@ -1,44 +1,46 @@
...
+        unauthorized,
+        /// The request or operation timed out.
```
danbugs, line 13:
```suggestion
        /// The request or operation timed out
```
~nit

wit/consumer.wit
danbugs: +1 for `subscribe` over `set-subscriptions`.

wit/request-reply.wit
danbugs: Currently, `request-options` only has two functions:
- `set-timeout-ms`, and
- `set-expected-replies`.

I understand `set-expected-replies` isn't used by `request` and only `request-multi`, but I feel like adding multiple structures could just add some confusion.

Do you feel like there are fields missing that would warrant the separation of `request-options` and `request-multi-options`? Otherwise, I'd opt to keeping them the same 👍

yordis: My concern is not "now" but "tomorrow." The options have two different use cases, and a copy-and-paste today could avoid a major breaking change in the future when configurations collide.

Cohesion around the use-case will avoid problematic global configuration options that the user needs to learn how to configure properly per use case.

danbugs:
> Please note that I just pushed one more commit (so we can pop it off if people hate it) to change how `request-multi` works. One of the uses of `request-multi` is to support a scatter/gather operation. In these cases, you might not know how many requests you are going to receive, so you can't set expected replies. Generally these wait until timeout and then return the results. This commit adds the ability to support all the different use cases for request-multi by adding expected replies to the request options. We don't have to support this, but when I was reviewing different use cases one more time, I was reminded of the scatter/gather operation. Just let me know if people hate it or if it isn't documented clearly

Btw - I am pretty sure that some implementors (like, maybe, Apache Kafka) do not inherently support a singular request, and, instead, often operate w/ returning a batch of messages as they are designed for high throughput. I think it's possible that you can still hack together a singular request implementation, but it's worth keeping in mind that that is a anti-pattern for some implementors.

wit/request-reply.wit
danbugs: At the same time, copy-and-paste here would mean that a host would have to copy-and-paste their implementation for, say, `set-timeout-ms`. This could lead to divergence in implementations that should be the same and hurt maintainability. That said, I understand what you're saying - maybe there's a middleground somewhere here...

yordis: Yeah, there's no question about your concern here.

Overall, let the implementor decide how much code-reusing they want to do and how the code path execution may look.

I am purely focusing on the specification detail.

thomastaylor312:
@danbugs So the more I look at this, the more I think having request and request-multi is what causes the problem here. As it currently stands a `request-multi` with expected replies set to 1 is the exact same as `request`, so why not just simplify it to one thing. It also avoids any problems with breaking interfaces in minor versions and any issues with implementations like Kafka

wit/request-reply.wit
thomastaylor312: FWIW, I originally had the idea of an optional `reply-to` field than then the handler can just call the `producer.send` function to send a message back, but got push back on it initially. To me there seems to be 3 options:

1. Have this be in the `metadata` of the message and entirely external to the interface
2. Optional reply to field that a handler can use with `producer.send`
3. A reply-request interface

My opinions on the 3:

1. Bad idea because every implementor may or may not have it and the naming is not standardized
2. Nicer because it is named, but if you want to depend on it, the implementation might not provide it
3. Best because an implementation can indicate if it supports replying. This is why the details of how to reply are left up to said implementation rather than being a field on the message

danbugs:
> @danbugs So the more I look at this, the more I think having request and request-multi is what causes the problem here. As it currently stands a `request-multi` with expected replies set to 1 is the exact same as `request`, so why not just simplify it to one thing. It also avoids any problems with breaking interfaces in minor versions and any issues with implementations like Kafka

I am good with unifying `request-multi` and `request` 👍

yordis:
I prefer if the spec works for `n` most of the time instead of `1` or `n`

wit/types.wit
```diff
@@ -1,44 +1,46 @@
...
-    /// A message with a binary payload, a format specification, and decorative metadata.
-    record message {
-        data: list<u8>,
-        format: format-spec,
-        metadata: option<list<tuple<string, string>>>
+    /// A message with a binary payload and additional information
+    resource message {
+        constructor(topic: string, data: list<u8>);
+        /// The topic/subject/channel this message was received or should be sent on
+        topic: func() -> string;
```
jocrau, line 27: We use CloudEvents for our IIoT machine data. A main objective is to decouple the event data from the publishing layer. Therefore, the topic in our model is in the pub/sub layer, while the data and the content-type is part of the event. Usually, the event/message is created at a different time and place than the publishing step. Also, an event can be published to multiple topics after its instanciation.

yordis: I needed to clarify this, but I never said anything; I self-reply. Could it be an ambiguous language problem?

danbugs:
I am closing this PR in favour of #24 which intends to simplify the wasi-messaging interface and hopefully remove some points of contention.

cc: @thomastaylor312 , @lukewagner , @brooksmtownsend , @devigned , @yordis , @jocrau

PR #24, opened 2024-10-04
Simplifies wasi-messaging interface with feedback

danbugs:
This PR is a follow-up to #23 . Like it, this also incorporates various points of feedback from interface champions, users, and working group discussions. In addition, it also simplifies the wasi-messaging interface as per feedback in the prior PR. 

The main changes to review are:
- updated sections (e.g., introduction, goals, and portability) in the `README`.
- removal of the `consumer.wit`.
- refactor of `messaging-guest` as `incoming-handler`.
- addition of `request-reply.wit`.
- refactor of `types.wit` including creating concrete error types, and adding a message resource.


.idea/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Default ignored files
```
yordis: Should `.idea` directory be ignored?

danbugs: Oops, good catch—included it on accident. Removed in https://github.com/WebAssembly/wasi-messaging/pull/24/commits/f3658470dd6cc0dc07fd4b11f2e3bd474718f85a 👍

wit/types.wit
```diff
@@ -1,44 +1,43 @@
...
+        set-content-type: func(content-type: string);
+        /// An opaque blob of data
+        data: func() -> list<u8>;
+        /// Set the opaque blob of data for this message, discarding the old value
+        set-data: func(data: list<u8>);
+        /// Optional metadata (also called headers or attributes in some systems) attached to the
+        /// message
+        metadata: func() -> option<list<tuple<string, string>>>;
+        /// Add a new key-value pair to the metadata, overwriting any existing value for the same key
+        add-metadata: func(key: string, value: string);
```
yordis, line 45: Should `set-metadata` also be available to "reset" the entired metadata?

yordis (commented):
They are multiple files with `No newline at end of file` not sure if that should be fix, ideal `.editorconfig` and things like that would help with IDEs doing the right thing

wit/types.wit
```diff
@@ -1,44 +1,43 @@
...
+        /// Set the content-type describing the format of the data in the message. This is
+        /// sometimes described as the "format" type
+        set-content-type: func(content-type: string);
+        /// An opaque blob of data
+        data: func() -> list<u8>;
+        /// Set the opaque blob of data for this message, discarding the old value
+        set-data: func(data: list<u8>);
+        /// Optional metadata (also called headers or attributes in some systems) attached to the
+        /// message
+        metadata: func() -> option<list<tuple<string, string>>>;
```
yordis:
```suggestion
        metadata: func() -> option<metadata>;
```

I wonder if we should include the `metadata` type as well

wit/request-reply.wit
```diff
@@ -0,0 +1,47 @@
...
+    /// the message is sent to is up to the implementation. This allows for reply-to details to be
+    /// handled in the best way possible for the underlying messaging system.
+    /// 
+    /// Please note that this reply functionality is different than something like HTTP because there
+    /// are several use cases in which a reply might not be required for every message (so this would
+    /// be a noop). There are also cases when you might want to reply and then continue processing.
+    /// Additionally, you might want to reply to a message several times (such as providing an
+    /// update). So this function is allowed to be called multiple times, unlike something like HTTP
+    /// where the reply is sent and the connection is closed.
+    reply: func(reply-to: borrow<message>, reply: message) -> result<_, error>;
```
yordis:
```suggestion
    reply: func(reply-to: borrow<message>, msg: message) -> result<_, error>;
```

Reusing the same domain language

.gitignore
```diff
@@ -1 +1,2 @@
-.DS_Store
\ No newline at end of file
+.DS_Store
```
Mossaka: I am not a fan of keeping an OS specific or IDE specific gitignore file in the remote repo. Perhaps you can add these in your local config, but don't commit to the remote?

wit/messaging.wit
```diff
@@ -2,10 +2,19 @@ package wasi:messaging@0.2.0-draft;
...
-world messaging {
+world imports-request-reply {
 	include imports;
-	export messaging-guest;
-}
\ No newline at end of file
+	import request-reply;
+}
+
+world messaging-request-reply {
```
Mossaka, line 19: suggest: add docs for each world in this package.

wit/producer.wit
```diff
@@ -1,5 +1,7 @@
+/// The producer interface is used to send messages to a channel/topic.
 interface producer {
-    use messaging-types.{client, channel, message, error};
+    use types.{client, message, error};
     
-    send: func(c: client, ch: channel, m: list<message>) -> result<_, error>;
+    /// Sends the message using the given client.
+    send: func(c: borrow<client>, m: message) -> result<_, error>;
```
jocrau: The `topic` should be an argument of the `send` function. I would also make the argument names more revealing:

```
func(client: borrow<client>, topic: topic, message: message)
```

wit/types.wit
```diff
@@ -1,44 +1,43 @@
...
-        data: list<u8>,
-        format: format-spec,
-        metadata: option<list<tuple<string, string>>>
+    /// A message with a binary payload and additional information
+    resource message {
+        constructor(topic: string, data: list<u8>);
+        /// The topic/subject/channel this message was received or should be sent on
+        topic: func() -> string;
+        /// Set the topic/subject/channel this message should be sent on
+        set-topic: func(topic: string);
```
jocrau: The `topic` should not be required for the purpose of publishing or sending it. It can be informational (more like metadata).

While the notion of a single topic or subject maps to some messaging systems very well (e.g. NATS), it is not a good fit for others (e.g. RabbitMQ).

yordis: I needed to clarify this, but I never said anything; I self-reply. Could it be an ambiguous language problem?

Is the intent that such a Message is the very last data structure in the component, so you must say where it is going at that point?

wit/types.wit
devigned: Or, `set-metadata: func(metadata: option<list<tuple<string, string>>>)` could replace `add-metadata`

wit/types.wit
jocrau: Just to clarify: I am advocating to add the `topic` to the `send` function, instead of the message. I believe that the topic is part of the publishing process, and not the message. The message should be able to stand by itself.

That said, keeping the `topic` in the message makes sense in the context of provenance.

wit/types.wit
danbugs: Addressed ✅

wit/request-reply.wit
danbugs: Addressed ✅

.gitignore
danbugs: Addressed ✅

wit/messaging.wit
danbugs: Addressed ✅

wit/types.wit
yordis: yeah replace, reset, set ... ambiguous language, I should use replace word 😄

wit/producer.wit
danbugs: Addressed making the argument names more revealing across all functions ✅ —as for the topic name being part of the parameter, what if we then have a discrepancy between the message's topic and the provided topic? I feel like this adds ambiguity for implementations that could hurt the portability of the interface.

wit/guest.wit
```diff
@@ -1,10 +1,15 @@
...
+    /// own decisions on how to handle errors returned from this function.
+    handle: func(ms: message) -> result<_, error>;
 
-    /// Whenever this guest receives a message in one of the subscribed channels, the message is sent to this handler
-    handler: func(ms: list<message>) -> result<_, error>;
+    // Subscribe to a list of topics (represented as `string`s) at runtime.
+    // Implementors should consider also allowing subscriptions to be made at compile time via
+    // some sort of configuration file. This function is intended to be called at the start of the
+    // guest's lifecycle before any messages are sent.
+    get-subscriptions: func() -> result<list<string>, error>;
```
yordis:
```suggestion
    // Returns the list of topics (represented as `string`s) at runtime.
    // Implementors should consider also allowing subscriptions to be made at compile time via
    // some sort of configuration file. This function is intended to be called at the start of the
    // guest's lifecycle before any messages are sent.
    get-topics: func() -> result<list<string>, error>;
```

Subscription word here is confusing; the language has been established around `topic` as far as I can tell.

wit/types.wit
danbugs: I'll defer this to @thomastaylor312 —do you want to elaborate a bit more on the reason behind the change of having topic be a part of `message` instead of an argument to `send`?

wit/guest.wit
danbugs: Addressed ✅

wit/types.wit
danbugs: Added a `remove-metadata` function that should address this use-case ✅

wit/types.wit
yordis: I thought that,
To make a point, I see `message` as a `send-command` single argument object, so such a command object has all the information for a given use case (send) to be done.

I know it is confusing because `message` is such a generic word, and the perspective of such an object matters; I am with the `topic` to be part of it because such an object is already the last object to be sent as a command message.

wit/types.wit
```diff
@@ -1,44 +1,51 @@
...
+        set-content-type: func(content-type: string);
+        /// An opaque blob of data
+        data: func() -> list<u8>;
+        /// Set the opaque blob of data for this message, discarding the old value
+        set-data: func(data: list<u8>);
+        /// Optional metadata (also called headers or attributes in some systems) attached to the
+        /// message
+        metadata: func() -> option<metadata>;
+        /// Add a new key-value pair to the metadata, overwriting any existing value for the same key
+        add-metadata: func(key: string, value: string);
```
yordis, line 45:
```suggestion
        /// Add a new key-value pair to the metadata, overwriting any existing value for the same key
        add-metadata: func(key: string, value: string);
        /// Set the metadata
        set-metadata: func(meta: metadata);
```

@danbugs, although, yes, we should keep `remove-metadata.` I was thinking about setting all the metadata when I do not know the keys or do not want to list the metadata and remove them individually.
Primarily thinking about relaying components that can not proxy the metadata at all

wit/producer.wit
jocrau: I'd argue for moving the `topic` from the `message` to the `send` function from two different sides: the domain modeling and the implementation side.

On the domain model side, a message should be independent from the (possibly many) topics it is published on. I should be able to send the message "Fire!" to `floors.first`, `floors.second`, and `floors.third` without changing or cloning it.

On the implementation side, backend systems have different semantics:

Kafka most closely resembles the original intent but introduces a `ProducerRecord` that is very closely tied to the semantics of the producer (in contrast to a message):

```
ProducerRecord<String, String> record = new ProducerRecord<>("floors.first", "Fire!");
producer.send(record);
```

RabbitMQ (`routing_key` as topic):

```
channel.basic_publish(exchange='alarms',
                      routing_key='floors.first',
                      body=message)
```

NATS (Rust client; subject = topic)

```
let subject = "floors.first";
let data = Bytes::from("Fire!");
client.publish(subject, data.clone()).await?;
```

yordis: > without changing or cloning it.

I think that is where the disconnection is happening, should they all be the `message`? Or should you have `n` message, one for each `send` if you know the topic is different?

wit/types.wit
thomastaylor312: Yep, I think @jocrau is correct here. I think the best course would be to remove `set-topic` from the message resource and then have topic be an argument on `send`

yordis: Does that mean anything related to `topic` would be removed from the `message` then?

danbugs: I think it is still good to have the getter for topic (for usage w/ `handle`), but it should not be required in the constructor anymore.

wit/types.wit
danbugs: Addressed ✅

wit/types.wit
danbugs: Added `topic` to `send`—`request-reply` pattern is untouched. You can still get `topic` metadata from `message` but no longer call `set-topic` on a message resource. 

Addressed ✅

wit/producer.wit
danbugs: Addressed ✅ —Added `topic` to send, @jocrau

danbugs:
This PR has been marinating for a while. So, if there's no additional feedback, I am hoping to merge this by Wednesday to maybe discuss moving the interface to Phase 2 on Thursday's WASI Subgroup video meeting.

wit/types.wit
yordis: My apologies, @danbugs. What is the resolution for resetting the metadata? Do I have to list every key and call `remove-metadata`?

danbugs: @yordis ~ I added the `set-metadata` method you suggested, so you could reset w/ an empty metadata list.

yordis: 🤦🏻 🤦🏻 🤦🏻 🤦🏻 🤦🏻 🤦🏻 gosh, ignore my stupidity sorry!

wit/producer.wit
```diff
@@ -1,5 +1,7 @@
+/// The producer interface is used to send messages to a channel/topic.
 interface producer {
-    use messaging-types.{client, channel, message, error};
+    use types.{client, message, error, topic};
     
-    send: func(c: client, ch: channel, m: list<message>) -> result<_, error>;
+    /// Sends the message using the given client.
+    send: func(c: borrow<client>, message: message, topic: topic) -> result<_, error>;
```
yordis:
```suggestion
    send: func(c: borrow<client>, topic: topic, message: message) -> result<_, error>;
```

Maybe? it

danbugs: Addressed ✅

wit/types.wit
danbugs: No, no worries 😆—always best to make sure!

yordis (approved):

PR #27, opened 2024-11-06
Adressing feedback from upstream sync

danbugs:
- clarified comment on `topic` function.
- added portability comment onto `metadata` to avoid it being used as an escape-hatch.
- removed `get-topics` func in favour of using wasi-config for runtime configuration.


thomastaylor312 (approved):
Thanks for pulling in all this feedback!

PR #29, opened 2024-12-05
refactor: make message property setters fallible

rvolosatovs:
Support for `content-type` and metadata/headers across different message queues and services is not uniform. For example, [RabbitMQ has built-in support for `content-type`](https://www.rabbitmq.com/docs/consumers#message-properties), but NATS.io does not.

Make setters of these properties on the message resource fallible to account for this difference.

Note, that this is only a partial solution to the bigger issue: What's the correct behavior on the host's part when content-type and/or metadata is set in the guest message, but not supported by the underlying implementation? (https://github.com/WebAssembly/wasi-messaging/issues/30)

### DIFF
--- a/imports-request-reply.md
+++ b/imports-request-reply.md
@@ -42,6 +42,19 @@ This allows the component to perform request/reply messaging patterns.</p>
 <p>A catch all for other types of errors
 </li>
 </ul>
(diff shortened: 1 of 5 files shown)


Mossaka:
Could you please take a look at this PR, @danbugs ?

danbugs (approved):

